PR #920, opened 2023-09-04
fix(gas-oracle): fetch base fee from the latest L1 block

colinlyguo:

Previously, the gas oracle sender utilized the base fee of the most recently finalized block to update the L2 gas oracle's recommended L1 gas price. This method caused a delay of at least 12.8 minutes, which is equivalent to 2 epochs.

In this PR, we have optimized the process to reduce this delay.

Key Changes:
1. The L1 watcher now extracts the base fee from the most recent L1 block.
2. When reorgs, soft-delete the outdated block and insert the new one.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes


database/migrate/migrations/00012_drop_unique_l1_block_number.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+drop index if exists l1_block_number_uindex;
```
0xmountaintop: I think we don't need to remove it. If you really want to handle reorg, you should add the complete handling logic like this:

+ soft delete from l1_block where index>=XXX
+ insert new l1_blocks
+ remember to do it inside a DB transaction

rollup/cmd/gas_oracle/app/app.go
```diff
@@ -65,35 +64,36 @@ func action(ctx *cli.Context) error {
...
 	if err != nil {
 		log.Error("failed to create new l2 relayer", "config file", cfgFile, "error", err)
 		return err
 	}
 	// Start l1 watcher process
 	go utils.LoopWithContext(subCtx, 10*time.Second, func(ctx context.Context) {
-		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, cfg.L1Config.Confirmations)
-		if loopErr != nil {
+		// Fetch the latest block number to decrease the delay when fetching gas prices
+		number, loopErr := l1Client.BlockNumber(ctx)
```
0xmountaintop: I think @zimpha mentions that we can use 1 or 10 confirmations? Why using latest here?

rollup/cmd/gas_oracle/app/app.go
```diff
@@ -65,35 +64,36 @@ func action(ctx *cli.Context) error {
...
+		if err != nil {
```
0xmountaintop: why change `loopErr` to `err`?

colinlyguo: It's a typo bug, fixed in [c83d6bd](https://github.com/scroll-tech/scroll/pull/920/commits/c83d6bd5d28cb5c925e8adc27cfbfefb29654b6e)

mask-pp: I don't recommend using the latest block height, how about using an independent confirm field in the config file?

rollup/cmd/gas_oracle/app/app.go
colinlyguo: I think using BlockNumber can achieve lower latency, and 1 to 10 confirmations also need to handle reorg, just in case.

database/migrate/migrations/00012_drop_unique_l1_block_number.sql
colinlyguo: adding reorg handling logic may be an overkill here because the table is only used for storing base fees now. and storing all fetched blocks (reorg blocks included) in db may be easier for debugging?

rollup/cmd/gas_oracle/app/app.go
colinlyguo: fetching L1 base bee can be a light-weight op.
btw, fetching the latest blocks seems to be better than other larger confirmation numbers, thus no need to add an extra config.

rollup/cmd/gas_oracle/app/app.go
zimpha: It is very uncommon that the reorg depth exceed `1`. Using `1` as confirmation should be safe.

rollup/cmd/gas_oracle/app/app.go
```diff
@@ -92,7 +92,9 @@ func action(ctx *cli.Context) error {
 	}
 	// Start l1 watcher process
 	go utils.LoopWithContext(subCtx, 10*time.Second, func(ctx context.Context) {
-		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, cfg.L1Config.Confirmations)
+		// Fetch the latest block number to decrease the delay when fetching gas prices
+		// Use latest block number - 1 to prevent frequent reorg
+		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, -2)
```
Thegaram:
```suggestion
		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, 2)
```

rollup/cmd/gas_oracle/app/app.go
colinlyguo: got it. fixed in [c83d6bd](https://github.com/scroll-tech/scroll/pull/920/commits/c83d6bd5d28cb5c925e8adc27cfbfefb29654b6e)

database/migrate/migrations/00012_drop_unique_l1_block_number.sql
colinlyguo: added in [c83d6bd](https://github.com/scroll-tech/scroll/pull/920/commits/c83d6bd5d28cb5c925e8adc27cfbfefb29654b6e)

georgehao (dismissed):

georgehao (approved):

rollup/cmd/gas_oracle/app/app.go
```diff
@@ -92,7 +92,9 @@ func action(ctx *cli.Context) error {
 	}
 	// Start l1 watcher process
 	go utils.LoopWithContext(subCtx, 10*time.Second, func(ctx context.Context) {
-		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, cfg.L1Config.Confirmations)
+		// Fetch the latest block number to decrease the delay when fetching gas prices
+		// Use latest block number - 1 to prevent frequent reorg
+		number, loopErr := butils.GetLatestConfirmedBlockNumber(ctx, l1client, 2)
```
zimpha: why use `2` here? It should be `rpc.LatestBlockNumber` and use `loopErr = l1watcher.FetchBlockHeader(number-1);` in the following lines.

colinlyguo: ~~In the implementation of `GetLatestConfirmedBlockNumber`, it's equivalent to `rpc.LatestBlockNumber` then `loopErr = l1watcher.FetchBlockHeader(number-1);`.~~

colinlyguo: fixed in [e66b5c6](https://github.com/scroll-tech/scroll/pull/920/commits/e66b5c6d3a1c0a6a9531b29aa8089cd2c4a8531e)

georgehao (approved):

zimpha (approved):

PR #981, opened 2023-10-08
fix(rollup-relayer): better handle commit batch revert

colinlyguo:

**Problem:**
When `rollup-relayer` encounters a batch commit failure, it continues to commit subsequent batches, which would fail.

**Solutions:**
1. Before committing a new batch, check If the parent batch has failed, if so, stop.
2. Use `estimateGas` for batches that have been committed failed.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes

### DIFF
--- a/rollup/internal/controller/relayer/l2_relayer.go
+++ b/rollup/internal/controller/relayer/l2_relayer.go
@@ -380,7 +380,7 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		if types.RollupStatus(batch.RollupStatus) == types.RollupCommitFailed {
 			// use eth_estimateGas if this batch has been committed failed.
 			fallbackGasLimit = 0
-			log.Warn("Batch commit previously failed, using eth_estimateGas for next submission", "hash", batch.Hash)
+			log.Warn("Batch commit previously failed, using eth_estimateGas for the re-submission", "hash", batch.Hash)
 		}
 		txHash, err := r.commitSender.SendTransaction(txID, &r.cfg.RollupContractAddress, big.NewInt(0), calldata, fallbackGasLimit)
 		if err != nil {
--- a/rollup/internal/controller/relayer/l2_relayer.go
+++ b/rollup/internal/controller/relayer/l2_relayer.go
@@ -378,8 +378,9 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		txID := batch.Hash + "-commit"
 		fallbackGasLimit := uint64(float64(batch.TotalL1CommitGas) * r.cfg.L1CommitGasLimitMultiplier)
 		if types.RollupStatus(batch.RollupStatus) == types.RollupCommitFailed {
-			// use estimateGas if this batch has been committed failed.
+			// use eth_estimateGas if this batch has been committed failed.
 			fallbackGasLimit = 0
+			log.Warn("Batch commit previously failed, using eth_estimateGas for next submission", "hash", batch.Hash)
 		}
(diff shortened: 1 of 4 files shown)


rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -371,6 +377,10 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		// send transaction
 		txID := batch.Hash + "-commit"
 		fallbackGasLimit := uint64(float64(batch.TotalL1CommitGas) * r.cfg.L1CommitGasLimitMultiplier)
+		if types.RollupStatus(batch.RollupStatus) == types.RollupCommitFailed {
+			// use estimateGas if this batch has been committed failed.
```
Thegaram: Add a `Warn` log here.

Thegaram (dismissed):

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: added in [74f93df](https://github.com/scroll-tech/scroll/pull/981/commits/74f93dfa41230e511f16bdca580ea5e43e008b45)

Thegaram (dismissed):

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -327,6 +327,12 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 				log.Error("Failed to get parent batch header", "index", batch.Index-1, "error", err)
 				return
 			}
+
+			if types.RollupStatus(parentBatch.RollupStatus) == types.RollupCommitFailed {
```
zimpha, line 331: what if the parent batch is going to failed but not failed yet? The rollup status will be submitted.

Thegaram: We will have multiple failing commit txs, the first tx fails with some reason, and the others fail with `incorrect parent batch hash`. But at some point, this line short-circuits the process, and we won't submit any more failing commit txs.

Also, during normal operation, there will only be 1 new batch / 2-4 minutes, so this mechanism will work well there.

0xmountaintop: > what if the parent batch is going to failed but not failed yet? The rollup status will be submitted.

make sense, but I think it;s acceptable.
 
 we will still submit this batch,
 we only get 5 batches each time
 next time it will probably already become failed.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -371,6 +377,11 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		// send transaction
 		txID := batch.Hash + "-commit"
 		fallbackGasLimit := uint64(float64(batch.TotalL1CommitGas) * r.cfg.L1CommitGasLimitMultiplier)
+		if types.RollupStatus(batch.RollupStatus) == types.RollupCommitFailed {
+			// use eth_estimateGas if this batch has been committed failed.
+			fallbackGasLimit = 0
+			log.Warn("Batch commit previously failed, using eth_estimateGas for next submission", "hash", batch.Hash)
```
0xmountaintop:
```suggestion
			log.Warn("Batch commit previously failed, using eth_estimateGas for the re-submission", "hash", batch.Hash)
```

0xmountaintop (dismissed):
just a nickpit

Thegaram (approved):

zimpha (approved):

PR #984, opened 2023-10-10
fix(gas-oracle): fetch block header

colinlyguo:

Change the outdated `range get` semantic to only get `blockHeight` L1 block.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes

### DIFF
--- a/rollup/internal/controller/watcher/l1_watcher.go
+++ b/rollup/internal/controller/watcher/l1_watcher.go
@@ -2,6 +2,7 @@ package watcher
 
 import (
 	"context"
+	"errors"
 	"math/big"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -122,6 +123,11 @@ func (w *L1WatcherClient) FetchBlockHeader(blockHeight uint64) error {
 	if err != nil {
 		log.Warn("Failed to get block", "height", blockHeight, "err", err)
 		return err
+	}
+
+	if block == nil {
+		log.Warn("Received nil block", "height", blockHeight)
+		return errors.New("received nil block")
 	}
 
 	var baseFee uint64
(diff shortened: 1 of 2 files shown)


0xmountaintop (approved):

zimpha (approved):

rollup/internal/controller/watcher/l1_watcher.go
```diff
@@ -116,52 +117,39 @@ func (w *L1WatcherClient) SetConfirmations(confirmations rpc.BlockNumber) {
...
+		GasOracleStatus: int16(types.GasOraclePending),
 	}
-	toBlock = height - 1
 
-	// insert succeed blocks
-	err = w.l1BlockOrm.InsertL1Blocks(w.ctx, blocks)
+	err = w.l1BlockOrm.InsertL1Blocks(w.ctx, []orm.L1Block{l1Block})
 	if err != nil {
-		log.Warn("Failed to insert L1 block to db", "fromBlock", fromBlock, "toBlock", toBlock, "err", err)
+		log.Warn("Failed to insert L1 block to db", "blockHeight", blockHeight, "err", err)
```
Thegaram, line 147: So if there's a reorg and we fetch the same block twice, we'll fail with this warning here? (If yes then that's OK)

Thegaram (approved):

rollup/internal/controller/watcher/l1_watcher.go
colinlyguo: Yes. I think based on this it would insert failed:
```
create unique index l1_block_number_uindex
on l1_block (number) where deleted_at IS NULL;
```

PR #1003, opened 2023-10-31
feat(contracts): remove proof verification on `finalizeBatchWithProof`

zimpha:

We won't provide the proof generation service for sepolia testnet. This PR will allow `finalizeBatchWithProof` without providing real proof.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

### Breaking change label

### DIFF
--- a/contracts/src/L1/rollup/ScrollChain.sol
+++ b/contracts/src/L1/rollup/ScrollChain.sol
@@ -292,7 +292,7 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         bytes32 _prevStateRoot,
         bytes32 _postStateRoot,
         bytes32 _withdrawRoot,
-        bytes calldata
+        bytes calldata _aggrProof
     ) external override OnlyProver whenNotPaused {
         require(_prevStateRoot != bytes32(0), "previous state root is zero");
         require(_postStateRoot != bytes32(0), "new state root is zero");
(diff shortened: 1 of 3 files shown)


contracts/src/L1/rollup/ScrollChain.sol
```diff
@@ -292,15 +292,15 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         bytes32 _prevStateRoot,
         bytes32 _postStateRoot,
         bytes32 _withdrawRoot,
-        bytes calldata _aggrProof
+        bytes calldata
```
icemelon: we should add an additional method that can finalize without proof on Sepolia. We will still finalize with proofs for some but not all batches.

zimpha: Then we also need to change rollup relayer, right?

colinlyguo (approved):

contracts/src/L1/rollup/ScrollChain.sol
```diff
@@ -353,6 +353,63 @@ contract ScrollChain is OwnableUpgradeable, PausableUpgradeable, IScrollChain {
         emit FinalizeBatch(_batchIndex, _batchHash, _postStateRoot, _withdrawRoot);
     }
 
+    /// @inheritdoc IScrollChain
+    function finalizeBatch(
```
colinlyguo, line 357: Could we add a `block.chainid != 1` check to double protect that the method cannot be called in Mainnet? Although in normal cases this version would not be deployed. Just in case.

icemelon: good point. Let's add this check

zimpha: This implementation never deployed to mainnet, it is not needed.

colinlyguo: Yes. Thus I mentioned that it's just a "double protection", to make this never happen even in some unexpected cases (e.g., an upgrade after a long time which mistakenly includes this function).

0xmountaintop (approved):

contracts/src/L1/rollup/ScrollChain.sol
icemelon: Agree with @colinlyguo. The additional check just prevents this being accidentally deployed to mainnet though it shouldn't happen

PR #1028, opened 2023-12-08
feature: update Dockerfile and githubAction

johnsonjie:
update Dockerfile and githubAction:

update Dockerfile, add a united workdir `/app` for all images, will be much more convenient when mounting config at starting.
Add `latest` tag when image builds, so we can both specify the tag or use the latest.


georgehao (dismissed):

colinlyguo (dismissed):

colinlyguo (approved):

georgehao (approved):

sbaizet-ledger (approved):
LGTM

PR #1037, opened 2023-12-08
fix(stack.go): return error when stack initialization fails

againest1:


prover/store/stack.go
```diff
@@ -43,6 +43,7 @@ func NewStack(path string) (*Stack, error) {
 	})
 	if err != nil {
 		log.Crit("init stack failed", "error", err)
+		return nil, err
```
0xmountaintop, line 46: `log.Crit` will terminate the whole process, the `return nil, err` will never be run

PR #1088, opened 2024-01-22
fix(bridge-history): incorrect status update and duplicated failed txs

colinlyguo:

This PR fixes several bugs:
1. failed relayed tx is not a terminal status, `_to.call` can fail with `OOG` and emit `FailedRelayedMessage`, in this case, retry will help.
2. remove db transaction because (i) all updates are idempotent. (ii) if during a transaction, the status is updated as "relayed" or "drop", the status would be incorrectly reset to "sent" in `replayMessage` scenario.
3. tweak `InsertFailedGatewayRouterTxs` to avoid inserting same transaction multiple times, which is very unfriendly to pagination.

Bridge cost optimization incompatibility:
5. Fix 1: EOA -> gateway router case.
6. Fix 2: EOA -> multisig -> gateway router case.
```
if tx.To() != nil && (*tx.To()).String() != e.cfg.GatewayRouterAddr { // EOA -> multisig -> gateway router.
	from = (*tx.To()).String()
} else { // EOA -> gateway router.
	signer := types.LatestSignerForChainID(new(big.Int).SetUint64(tx.ChainId().Uint64()))
	sender, senderErr := signer.Sender(tx)
	if senderErr != nil {
		log.Error("get sender failed", "chain id", tx.ChainId().Uint64(), "tx hash", tx.Hash().String(), "err", senderErr)
		return nil, nil, senderErr
	}
	from = sender.String()
}
```

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [x] Yes

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206415608869443


georgehao (dismissed):

bridge-history-api/internal/logic/l2_event_parser.go
```diff
@@ -113,12 +123,31 @@ func (e *L2EventParser) ParseL2EventLogs(logs []types.Log, blockTimestampsMap ma
 			event := backendabi.L2SentMessageEvent{}
 			err := utils.UnpackLog(backendabi.IL2ScrollMessengerABI, &event, "SentMessage", vlog)
 			if err != nil {
-				log.Warn("Failed to unpack SentMessage event", "err", err)
+				log.Error("Failed to unpack SentMessage event", "err", err)
 				return nil, nil, err
 			}
+			from := event.Sender.String()
```
zimpha: the code looks the same in l1 fetcher, could use function to wrap the code?

colinlyguo: make sense. fixed in [07426cf](https://github.com/scroll-tech/scroll/pull/1088/commits/07426cff4d7afb357c5d13aa1e4b7e7a8136e87d)

zimpha (approved):

georgehao (approved):

PR #1011, opened 2023-11-09
feat(contracts): bridge costs optimization

zimpha:
### Purpose or design rationale of this PR

This PR try to reduce the gas costs for bridging. The following changes are made:

- change `counterpart`, `router`, `messenger` in gateway from storage variable to immutable variable
- change `rollup` and `messageQueue` in `L1ScrollMessenger` from storage variable to immutable variable
- change `messageQueue` in `ScrollChain` from storage variable to immutable variable
- change `messenger` and `scrollChain` in `L1MessageQueue` from storage variable to immutable variable
- add `L1MessageQueueWithGasPriceOracle` to merge `L2GasPriceOracle` and `L1MessageQueue`

### PR title

Your PR title must follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) (as we are doing squash merge for each PR), so it must start with one of the following [types](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type):

- [x] perf: A code change that improves performance
- [x] refactor: A code change that doesn't fix a bug, or add a feature, or improves performance

### Deployment tag versioning

Has `tag` in `common/version.go` been updated or have you added `bump-version` label to this PR?

- [x] No, this PR doesn't involve a new deployment, git tag, docker image tag

### Breaking change label

Does this PR have the `breaking-change` label?

- [x] No, this PR is not a breaking change

PR #1093, opened 2024-01-29
fix(contracts): OZ-H01 ETH Deposits Can Get Stuck if They Are Not Successfully Bridged

zimpha:

This PR fix the following issue reported by Openzepplin: **H-01 ETH Deposits Can Get Stuck if They Are Not Successfully Bridged**

> [Pull request #1011](https://github.com/scroll-tech/scroll/pull/1011) introduced the change of [redirecting the calls](https://github.com/scroll-tech/scroll/pull/1011/files#diff-ef2695307fe7f3dfd610ae22338941d20233e19db326322cfca61ad4ceeda1c7) to deposit ETH from the L1GatewayRouter contract to the L1ScrollMessenger contract without going through the L1ETHGateway contract. This was done with the intention of reducing the gas cost associated with such an action.
> 
> However, this causes a problem. Namely, in situations in which a message could not be correctly sent through the bridge, the [dropping and asset-return mechanism](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/L1ScrollMessenger.sol#L262-L311) implemented in the L1ScrollMessenger contract will get stuck and the assets will not be able to be paid back. This is due to the lack of the [onDropMessage hook](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/L1ScrollMessenger.sol#L308) implementation in the [L1GatewayRouter contract](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol), which serves as a handler to repay the respective origin of the message.
>
> For instance, if a user wants to simply deposit ETH, in both versions they should call the [depositETH function](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol#L206) from the L1GatewayRouter contract. The difference lies in [who calls](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/L1ScrollMessenger.sol#L340) and passes the message to the L1ScrollMessenger contract. In the former implementation, the [message would come from](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1ETHGateway.sol#L148) the L1ETHGateway contract, whereas in the current implementation, the L1GatewayRouter will be the [caller](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol#L226). This is relevant as the _msgSender call will then be part of the [_xDomainCalldata](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/L1ScrollMessenger.sol#L340) data that will be used to keep track of the message (with its hash) but will also be used in case the message needs [to be dropped](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/L1ScrollMessenger.sol#L308).
>
> In such a dropping scenario, as the address that sent the message to the L1ScrollMessenger contract is the one that will be called to execute the onDropMessage hook, if such hook is not implemented, the dropping mechanism will fail and the original user will not get their ETH back. This is the same as how it used to happen when routing the call through the L1ETHGateway contract. As this does not depend on the data added to the depositETH call, those funds will get stuck in case they are not bridged successfully.
> 
> To showcase this, one can get inspired by the following [gist](https://gist.github.com/andresbach/5dcd4fff562dc6c72ac713e46136bd27); however, caution should be made when fixing this issue, since the gist's proposed scenario in which the issue resolves is merely an example and it is not meant to represent a fully valid resolution.
>
>Consider implementing the onDropMessage hook in the L1GatewayRouter contract to handle the back payment when dropping messages.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes


contracts/src/L1/gateways/L1GatewayRouter.sol
```diff
@@ -140,6 +163,14 @@ contract L1GatewayRouter is OwnableUpgradeable, IL1GatewayRouter {
         return _amount;
     }
 
+    /// @inheritdoc IMessageDropCallback
+    function onDropMessage(bytes calldata _message) external payable virtual onlyInDropContext {
```
Thegaram: Any concern for reentrancy here? (I assume no since `dropMessage` has `notInExecution`.)

contracts/src/L1/gateways/L1GatewayRouter.sol
```diff
@@ -223,7 +254,13 @@ contract L1GatewayRouter is OwnableUpgradeable, IL1GatewayRouter {
         bytes memory _data,
         uint256 _gasLimit
     ) public payable override onlyNotInContext {
-        IL1ScrollMessenger(messenger).sendMessage{value: msg.value}(_to, _amount, _data, _gasLimit, _msgSender());
+        IL1ScrollMessenger(messenger).sendMessage{value: msg.value}(
+            _to,
+            _amount,
+            abi.encode(_msgSender(), _data),
```
Thegaram: How will the receiver decode the data? Shouldn't that be done by the bridge, so that the receiver is called with `_data` and not with `abi.encode(_msgSender(), _data)`?

Thegaram: Discussed with Xi:

The behaviour of `gatewayRouter.depositETHAndCall` will slightly change. Previously, the target contract on L2 would receive a callback from `L2ETHGateway`: `IScrollGatewayCallback(_to).onScrollGatewayCallback(_data);`

Now, the target contract on L2 would directly be called by `L2ScrollMessenger` with the payload of `abi.encode(_msgSender(), _data).`

And analogously for withdrawals.

Thegaram (dismissed):

Thegaram (dismissed):

icemelon (dismissed):

icemelon (approved):

PR #1094, opened 2024-01-29
fix(contracts): OZ-L01 Implementation Keeps Functionalities for Deprecated Variables

zimpha:

This PR fix the following issue reported by Openzepplin: **L-01 Implementation Keeps Functionalities for Deprecated Variables**

> At line [82](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol#L82) of L1GatewayRouter.sol, it is explained that the ethGateway parameter is no longer in use. However, the logic that makes use of/changes this variable, such as the [check](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol#L96) in the initialize function and the [setETHGateway](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/gateways/L1GatewayRouter.sol#L244) function, is maintained.
> 
> If a variable is no longer in use, consider removing the logic that uses it.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes


0xmountaintop (dismissed):

Thegaram (dismissed):
I'd recommend removing `setETHGateway`. Other than that it looks good.

Thegaram (approved):

PR #1096, opened 2024-01-29
fix(contracts): OZ-N02 Renaming Opportunities

zimpha:

This PR fix the following issue reported by Openzepplin: **N-02 Renaming Opportunities**

>The [INTRINSIC_GAS_NONZERO_BYTE factor](https://github.com/scroll-tech/scroll/blob/45e1305de5f7a1dc5aadadd7d3bb6726a5dee949/contracts/src/L1/rollup/L1MessageQueueWithGasPriceOracle.sol#L21) that multiplies the length of the message-to-be-sent in order to calculate the amount of gas needed for the L2 execution, has a name that originates from a previous version that differentiated between zero and non-zero bytes. However, now, there is no such distinction and its name suggests that the whole message does not have a zero byte.
>
>Consider changing its name to something more appropriate that does not refer to old code's behavior.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes


0xmountaintop (approved):

Thegaram (approved):
Personally I'd prefer `INTRINSIC_GAS_NONZERO_BYTE` since that's how it's defined in the Ethereum specs, but this change is also acceptable.

PR #1103, opened 2024-01-31
fix(bridge-history): use msg.sender encoded in message field of SentMessage event

colinlyguo:

The `msg.sender` is added in: https://github.com/scroll-tech/scroll/pull/1093
Encode `msg.sender` in the contract: https://github.com/scroll-tech/scroll/blob/9d20b34c47c34275f393dc7b3e13c94759f92920/contracts/src/L1/gateways/L1GatewayRouter.sol#L260
Decode `msg.sender` in the contract: https://github.com/scroll-tech/scroll/blob/9d20b34c47c34275f393dc7b3e13c94759f92920/contracts/src/L1/gateways/L1GatewayRouter.sol#L169

This will keep the behavior consistent with previous version of `bridge-history-backend`.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label


georgehao (dismissed):

bridge-history-api/internal/logic/l1_event_parser.go
```diff
@@ -283,26 +283,19 @@ func (e *L1EventParser) ParseL1MessageQueueEventLogs(logs []types.Log, l1Deposit
...
-		}
-		return sender.String(), nil
+func getRealFromAddress(eventSender common.Address, eventMessage []byte, gatewayRouterAddr string) (string, error) {
+	if eventSender != common.HexToAddress(gatewayRouterAddr) {
+		return eventSender.String(), nil
 	}
-	return from, nil
+
+	// deposit/withdraw ETH: EOA -> contract 1 -> ... -> contract n -> gateway router -> messenger.
+	if len(eventMessage) < 32 {
```
zimpha: maybe we should not return error, since sepolia is running a version without such fix.

colinlyguo: aha, makes sense, I'll change it to a warning log.
currently, the sender could be the gateway router address in sepolia because the contract has been upgraded once.

colinlyguo: added back the legacy logic as a fallback, cannot handle 100% of cases before the next upgrade, but it's better than nothing.
commit: [9218957](https://github.com/scroll-tech/scroll/pull/1103/commits/9218957717869110924328803d8e6807cdb6630d)

colinlyguo: cannot handle the case that in current sepolia (if 1 and 2 happen simultaneously):
1. multiple contracts: deposit/withdraw ETH: EOA -> contract 1 -> ... -> contract n -> gateway router.
2. data field >= 32 bytes

zimpha (approved):

bridge-history-api/internal/logic/l1_event_parser.go
zimpha: yeah, let's hope no one has used the gateway router with data field >= 32 bytes.

georgehao (approved):

PR #1162, opened 2024-02-26
feat(contracts) : added functions

cryptoryda:
### Functions added to Calculate the maximum output amount of Ether for a given input token amount.


cryptoryda:
Not Needed No?

PR #1170, opened 2024-02-29
fix(contracts): allow setting ETH/token limit to zero

arularmstrong:
Previously, the updateLimit method in Ratelimiter enforced that the new token limit cannot be zero. However, there are scenarios where the admin may need to temporarily halt the process by setting the limit to zero. This commit modifies the updateLimit method to allow setting the token limit to zero when needed.


Thegaram:
Rate limiter is not used anymore.

arularmstrong:
can you atleast describe the active files on readme?

PR #1174, opened 2024-03-04
feat: make chunk and batch proposers respect fork boundaries

omerfirmak:
### PR title

- [X] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [X] No, this PR doesn't involve a new deployment, git tag, docker image tag
- [ ] Yes

### Breaking change label

- [ ] Yes


.github/workflows/bridge_history_api.yml
```diff
@@ -32,7 +32,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.20.x
+        go-version: 1.22.x
```
colinlyguo: Why upgrade the go-version to 1.22, Any considerations or just bumping the versions for robustness?

omerfirmak: mainly to get access to `slices` package, I can work around it if you would rather stay on go 1.20

colinlyguo: Got it. Just a quick question. I think we can keep it for now and wait for more reviews.

georgehao: > mainly to get access to slices package, I can work around it if you would rather stay on go 1.20

which function do you want to use ?

georgehao: Oh, the go latest version is 1.22.0, i think it's too new to use

common/network/network.go
```diff
@@ -0,0 +1,69 @@
...
+		config.ArchimedesBlock,
+		config.ShanghaiBlock,
+	} {
+		if fork == nil {
+			continue
+		} else if !fork.IsUint64() {
+			panic("fork set to a height larger than max uint64")
+		} else if height := fork.Uint64(); height != 0 {
+			forkHeights = append(forkHeights, height)
+		}
```
georgehao:
```suggestion
		forkHeights = append(forkHeights, fork.Uint64())
```

How about change to this. because the `[]*big.Int` is decided, so the exception check is redundancy, i think

common/network/network.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	slices.Sort(forkHeights)
+	return slices.Compact(forkHeights) // Remove duplicates
```
georgehao:
```suggestion
	return forkHeights
```

.github/workflows/bridge_history_api.yml
georgehao: the slices.Index is add in go1.21.0, can switch to this version

omerfirmak: Sure, will switch to 1.21 :+1:

common/network/network.go
omerfirmak: We need forkHeights to be sorted and have no duplicates.

georgehao:
```
for _, fork := range []*big.Int{
		config.HomesteadBlock,
		config.DAOForkBlock,
		config.EIP150Block,
		config.EIP155Block,
		config.EIP158Block,
		config.ByzantiumBlock,
		config.ConstantinopleBlock,
		config.PetersburgBlock,
		config.IstanbulBlock,
		config.MuirGlacierBlock,
		config.BerlinBlock,
		config.LondonBlock,
		config.ArrowGlacierBlock,
		config.ArchimedesBlock,
		config.ShanghaiBlock,
	}
```
This list is decided, and there is no duplicates in this list. So I think the checks is no necessary.

common/network/network.go
omerfirmak: Sure, if you say it is safe to assume fork heights to always fit in a `uint64`.

common/network/network.go
omerfirmak: > and there is no duplicates in this list

Different forks can be scheduled for the same block tho

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -253,7 +267,8 @@ func (p *BatchProposer) proposeBatchChunks() ([]*orm.Chunk, *types.BatchMeta, er
 
 	currentTimeSec := uint64(time.Now().Unix())
 	if dbChunks[0].StartBlockTime+p.batchTimeoutSec < currentTimeSec ||
-		totalChunks == p.maxChunkNumPerBatch {
+		totalChunks == p.maxChunkNumPerBatch ||
+		hitForkBoundary {
```
georgehao:
```suggestion
		totalChunks == p.maxChunkNumPerBatch || hitForkBoundary {
```

.github/workflows/bridge_history_api.yml
omerfirmak: Done

common/network/network.go
```diff
@@ -0,0 +1,77 @@
...
+	"fmt"
+	"math/big"
+	"sort"
+
+	"github.com/scroll-tech/go-ethereum/params"
+)
+
+// Network represents a known network
+type Network string
+
```
georgehao: As https://scrollco.slack.com/archives/C02PKAQ2430/p1709044228395759 discussed, maybe we reach a consensus：don't import the genesis.json for `rollup_relayer` and `coordinator`. 

And if really need this, coordinator will also use this. BTW, I think just add a config option to currently config.json will also  achieve the same effect. How do you think? @colinlyguo @Thegaram

omerfirmak: > maybe we reach a consensus：don't import the genesis.json for rollup_relayer and coordinator.

I believe, in that thread, the consensus was to use the genesis config, which is what this PR does.

georgehao: oh okay, maybe i don't express the opinion clearly in the thread. actually my mean that we don't need import another config file. And we just use one option from the genesis.json, besides I think the current option (like config.HomesteadBlock, config.DAOForkBlock and etc) can't satisfied, maybe add other option like `config.ScrollAprilBlock`. if so, why not import the option to config.json

georgehao: Synced with @omerfirmak , I think the way is make sense.

rollup/cmd/rollup_relayer/app/app.go
```diff
@@ -78,12 +79,17 @@ func action(ctx *cli.Context) error {
 		log.Crit("failed to create l2 relayer", "config file", cfgFile, "error", err)
 	}
 
-	chunkProposer := watcher.NewChunkProposer(subCtx, cfg.L2Config.ChunkProposerConfig, db, registry)
+	network := network.Network(ctx.String(utils.NetworkFlag.Name))
+	if !network.IsKnown() {
```
omerfirmak: Is there a need to use arbitrary genesis configurations in rollup relayer or just relying on the hard coded configurations from l2geth is fine? cc: @Thegaram

georgehao: walkthrough the code, if don't load the genesis.json, i think the block_number all is 0

https://github.com/scroll-tech/go-ethereum/blob/develop/params/config.go#L342

Thegaram: It's better if we can support custom configurations, we will need this for local and staging testnets.

common/network/network.go
```diff
@@ -0,0 +1,77 @@
...
+var (
+	// Mainnet network
+	Mainnet Network = "mainnet"
+	// Sepolia network
+	Sepolia Network = "sepolia"
+	// Alpha network
+	Alpha Network = "alpha"
```
Thegaram: We can remove Alpha, that network has been deprecated.

rollup/cmd/rollup_relayer/app/app.go
omerfirmak: > if we can support custom configurations

in that case, I will remove the `Network` enum and just pass in path to the `genesis.json` to be parsed by `rollup_relayer`. does it make sense to you @georgehao  @Thegaram ?

common/forks/forks.go
```diff
@@ -0,0 +1,45 @@
...
+		config.EIP158Block,
+		config.ByzantiumBlock,
+		config.ConstantinopleBlock,
+		config.PetersburgBlock,
+		config.IstanbulBlock,
+		config.MuirGlacierBlock,
+		config.BerlinBlock,
+		config.LondonBlock,
+		config.ArrowGlacierBlock,
+		config.ArchimedesBlock,
```
georgehao, line 27: Shall we define scroll's own hard fork block number? I think the ethereum hard fork block number isn't fit for us

omerfirmak: I believe that is out-of-scope for this PR, but whenever a new fork block configuration is added to `params.ChainConfig`, it should be added to this list as well.

georgehao: make sense

rollup/cmd/rollup_relayer/app/app.go
```diff
@@ -121,6 +131,19 @@ func action(ctx *cli.Context) error {
 	return nil
 }
 
+func readGenesis(genesisPath string) (*core.Genesis, error) {
+	file, err := os.Open(filepath.Clean(genesisPath))
+	if err != nil {
```
georgehao: How about move this logic to [here](https://github.com/scroll-tech/scroll/tree/develop/rollup/internal/config)

rollup/internal/controller/watcher/chunk_proposer.go
```diff
@@ -184,14 +190,31 @@ func (p *ChunkProposer) updateChunkInfoInDB(chunk *types.Chunk) error {
 	return err
 }
 
+// blocksUntilFork returns the number of blocks until the next fork
+// returns 0 if there is no fork scheduled for the future
+func blocksUntilFork(blockHeight uint64, forkHeights []uint64) uint64 {
```
georgehao:
```suggestion
func (p *ChunkProposer) blocksUntilFork(blockHeight uint64, forkHeights []uint64) uint64 {
```

common/forks/network_test.go
```diff
@@ -0,0 +1,26 @@
+package forks
```
georgehao: maybe change `network_test.go` to `forks_test.go` will be better

rollup/internal/controller/watcher/chunk_proposer_test.go
```diff
@@ -141,6 +145,18 @@ func testChunkProposerLimits(t *testing.T) {
...
 		},
+		{
+			name:                       "ForkBlockReached",
+			maxBlockNum:                100,
+			maxTxNum:                   10000,
+			maxL1CommitGas:             50000000000,
+			maxL1CommitCalldataSize:    1000000,
+			maxRowConsumption:          1000000,
+			chunkTimeoutSec:            1000000000000,
+			expectedChunksLen:          1,
```
georgehao, line 155: I think this unit test don't cover the hard fork logic.

rollup/internal/controller/watcher/batch_proposer_test.go
```diff
@@ -83,6 +86,16 @@ func testBatchProposerLimits(t *testing.T) {
 			expectedBatchesLen:         1,
 			expectedChunksInFirstBatch: 1,
 		},
+		{
+			name:                       "ForkBlockReached",
+			maxChunkNum:                10,
+			maxL1CommitGas:             50000000000,
+			maxL1CommitCalldataSize:    1000000,
+			batchTimeoutSec:            1000000000000,
```
georgehao, line 94: ditto

rollup/conf/genesis.json
```diff
@@ -0,0 +1,16 @@
+{
```
colinlyguo, line 1: This method enlightens me that maybe we can also reuse some other info (e.g., part of chunk-proposer and batch-proposer limits) from the genesis config.

common/forks/forks.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	var forkHeights []uint64
+	for height := range forkHeightsMap {
+		forkHeights = append(forkHeights, height)
+	}
+	sort.Slice(forkHeights, func(i, j int) bool {
+		return forkHeights[i] < forkHeights[j]
+	})
+	return forkHeights, forkHeightsMap
```
colinlyguo, line 44: `forkHeightsMap` is an unused return value.

rollup/internal/controller/watcher/chunk_proposer_test.go
```diff
@@ -141,6 +145,18 @@ func testChunkProposerLimits(t *testing.T) {
 			expectedChunksLen:          1,
 			expectedBlocksInFirstChunk: 1,
 		},
+		{
```
colinlyguo, line 147: can also add a multiple hard-fork test (although this would rarely happen).

colinlyguo (dismissed):
LGTM. Left some comments, and the refactor comments are the same as @georgehao.

georgehao:
BTW, I will checkout your branch to write coordinator logic: `assign task to multi version prover`, because I want reuse your `forks` common logic. Pls don't use `hard push`. Thx, hahaha

rollup/internal/controller/watcher/chunk_proposer.go
omerfirmak: I will move it to `forks`package, if that makes sense to you as well.

common/forks/forks.go
omerfirmak: https://github.com/scroll-tech/scroll/pull/1174/files/e608c6bb3edea37719f3a00af909c45ee0de2664#diff-988037fb2a3b2ec2f89c5a3e6be65cba29cf2f45410c8a5214284982fa1d73d9R70 batch proposer does

rollup/internal/controller/watcher/chunk_proposer.go
georgehao: yeah. make sense

rollup/internal/controller/watcher/chunk_proposer_test.go
omerfirmak: Can you elaborate? I am fairly certain they cover the changes. These tests fail without the changes I made in chunk and batch proposers.

common/forks/forks.go
colinlyguo: Oh, I see. Thanks.

rollup/internal/controller/watcher/chunk_proposer_test.go
omerfirmak: An additional test doesn't really provide additional coverage. Multiple hard-forks only matter for `BlocksUntilFork` which is properly tested [here](https://github.com/scroll-tech/scroll/pull/1174/files#diff-c614c84c333c0ed1b47977755aa8ff548f4263df17a46dc059af00c7645f4442R28)

rollup/internal/controller/watcher/chunk_proposer_test.go
georgehao: Yes, you are right.

chunk: 
image

batch:
image

georgehao (approved):

colinlyguo (approved):

PR #1182, opened 2024-03-07
feat(coordinator): support assigning tasks of different versions to provers

georgehao:

* assign prover task to multi version prover
* coordinator import genesis.json
* coordinator get_task add `fork_block_number` param
* prover add `fork_block_number` param

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes


coordinator/go.mod
bzarka, line 1: package app

import (
	"crypto/rand"
	"encoding/json"
	"fmt"
	"math/big"
	"os"
	"strconv"
	"testing"
	"time"

	"github.com/scroll-tech/go-ethereum/params"
	coordinatorConfig "scroll-tech/coordinator/internal/config"
	"scroll-tech/common/cmd"
	"scroll-tech/docker"
)

var (
	httpStartPort int64 = 30000
)

// CoordinatorApp coordinator-test client manager.
type CoordinatorApp struct {
	Config         *coordinatorConfig.Config
	ChainConfig    *params.ChainConfig
	base           *docker.App
	originFile     string
	configFile     string
	chainConfigFile string
	coordinatorFile string
	genesisFile    string
	HTTPPort       int64
	args           []string
	docker.AppAPI
}

// NewCoordinatorApp returns a new coordinatorApp manager.
func NewCoordinatorApp(base *docker.App, configFile string, chainConfigFile string) *CoordinatorApp {
	coordinatorFile := fmt.Sprintf("/tmp/%d_coordinator-config.json", base.Timestamp)
	genesisFile := fmt.Sprintf("/tmp/%d_genesis.json", base.Timestamp)
	port, _ := rand.Int(rand.Reader, big.NewInt(2000))
	httpPort := port.Int64() + httpStartPort
	coordinatorApp := &CoordinatorApp{
		base:           base,
		originFile:     configFile,
		configFile:     configFile,
		chainConfigFile: chainConfigFile,
		coordinatorFile: coordinatorFile,
		genesisFile:    genesisFile,
		HTTPPort:       httpPort,
		args:           []string{"--log.debug", "--config", coordinatorFile, "--genesis", genesisFile, "--http", "--http.port", strconv.Itoa(int(httpPort))},
	}
	if err := coordinatorApp.MockConfig(true); err != nil {
		panic(err)
	}
	return coordinatorApp
}

// HTTPEndpoint returns the HTTP endpoint of the coordinator.
func (c *CoordinatorApp) HTTPEndpoint() string {
	return fmt.Sprintf("http://localhost:%d", c.HTTPPort)
}

// MockConfig creates a new coordinator config.
func (c *CoordinatorApp) MockConfig(store bool) error {
	cfg, err := coordinatorConfig.NewConfig(c.configFile)
	if err != nil {
		return err
	}
	cfg.Auth.LoginExpireDurationSec = 1
	c.Config = cfg

	genesis, err := coordinatorConfig.ReadGenesis(c.chainConfigFile)
	if err != nil {
		return err
	}
	chainConf := genesis.Config
	chainConf.HomesteadBlock = big.NewInt(1)
	c.ChainConfig = chainConf

	if !store {
		return nil
	}

	coordinatorConfigData, err := json.Marshal(c.Config)
	if err != nil {
		return err
	}
	genesisConfigData, err := json.Marshal(genesis)
	if err != nil {
		return err
	}

	if writeErr := os.WriteFile(c.coordinatorFile, coordinatorConfigData, 0600); writeErr != nil {
		return writeErr
	}
	if writeErr := os.WriteFile(c.genesisFile, genesisConfigData, 0600); writeErr != nil {
		return writeErr
	}
	return nil
}

bzarka (commented):
Hi,

This refactored version of the code includes:

Consolidation of import statements to a single block for better readability.
Proper alignment and formatting of struct fields and their corresponding comments.
Updated function parameters to reflect the changes made in the constructor function NewCoordinatorApp.
Renamed variables originFile to configFile and HTTPPort to httpPort for consistency and clarity.
Replaced direct file paths with variables where applicable to enhance maintainability.
Improved error handling by returning specific error types where possible.
Reorganized the logic inside the MockConfig function for better readability and maintainability.

coordinator/internal/types/get_task.go
```diff
@@ -2,9 +2,10 @@ package types
 
 // GetTaskParameter for ProverTasks request parameter
 type GetTaskParameter struct {
-	ProverHeight int    `form:"prover_height" json:"prover_height"`
-	TaskType     int    `form:"task_type" json:"task_type"`
-	VK           string `form:"vk" json:"vk"`
+	ForkBlockNumber uint64 `form:"fork_block_number" json:"fork_block_number"`
```
Thegaram: What does `ForkBlockNumber` mean?

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,42 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
 		return nil, fmt.Errorf("check prover task parameter failed, error:%w", err)
 	}
 
+	// TODO Only happen when the first time hard fork, the ForkBlockNumber = 0.
+	// TODO need check fork block number whether equal 0 after the first hard fork
+	if getTaskParameter.ForkBlockNumber != 0 && !bp.forkMap[getTaskParameter.ForkBlockNumber] {
```
Thegaram: I'm not sure I understand the logic here.

The "fork block number" should not be reported by the prover. We know the prover version (e.g. v0 or v1), and we know the hard for block numbers, so the coordinator can decide which chunks/batches to assign to which provers.

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,46 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
 	}
 
+	var hardForkNumber uint64
+	if getTaskParameter.HardForkName != "" {
+		var exist bool
+		hardForkNumber, exist = bp.nameForkMap[getTaskParameter.HardForkName]
+		if !exist {
+			log.Debug("hard fork prover get empty batch because of the hard fork name don't exist", "height",
+				getTaskParameter.ProverHeight, "fork name", getTaskParameter.HardForkName)
+			return nil, nil
```
Thegaram: Why not return error?

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,46 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
 		return nil, fmt.Errorf("check prover task parameter failed, error:%w", err)
 	}
 
+	var hardForkNumber uint64
+	if getTaskParameter.HardForkName != "" {
+		var exist bool
+		hardForkNumber, exist = bp.nameForkMap[getTaskParameter.HardForkName]
+		if !exist {
+			log.Debug("hard fork prover get empty batch because of the hard fork name don't exist", "height",
```
Thegaram: Later we can add a metric to track this. And to track number of task requests by different fork.

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,46 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+		}
+	}
+
+	if getTaskParameter.HardForkName == "" {
+		hardForkNumber = bp.maxForkNumber - 1
+	}
+
+	// if the hard fork number set, rollup relayer must generate the chunk from hard fork number,
+	// so the hard fork chunk's start_block_number must be ForkBlockNumber
+	hardForkStartChunk, getStartChunkErr := bp.chunkOrm.GetChunkByStartBlockNumber(ctx, bp.maxForkNumber)
```
Thegaram: This works but it seems overcomplicated to me.

To me, the logical approach would be:
- Based on `hardForkNumber`, get hard fork block range. The results is e.g. `[0, 99]` and `[100, MAX]`.
- In db chunk query, add condition on block range

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: discussed

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: make sense

georgehao: fixed [32dc659](https://github.com/scroll-tech/scroll/pull/1182/commits/32dc6591862afe599ed881c73ad01156294e718e)

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: fixed [9399974](https://github.com/scroll-tech/scroll/pull/1182/commits/93999746c9612d06f0c99657255d385d5d6183c9)

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: `bp.batchOrm.GetAssignedBatch` query the batch table, only have `start_chunk_index` and `end_chunk_index` in the batch table. So if batch want know the `start_block_number` of batch, batch sql need inner join with chunk.

So I split the inner join sql.

Thegaram: What I don't like about this approach are:
- Special handing for the "last fork". If we spin up a test network and set forks to `100` and `200` it will only care about the last one.
- Passing `isFork` to the DB queries. I feel like the orm layer does not need to be aware of forks.

georgehao: make sense

georgehao: fixed [5da9b94](https://github.com/scroll-tech/scroll/pull/1182/commits/5da9b94d1bd0e1a3029a5f0f00b37d4d7b74b3be)

common/forks/forks_test.go
```diff
@@ -64,3 +69,64 @@ func TestBlocksUntilFork(t *testing.T) {
...
+			name:         "To200-1",
+			forkHeight:   100,
+			forkHeights:  []uint64{100, 200},
+			expectedFrom: 100,
+			expectedTo:   200,
+		},
+		{
+			name:         "to2",
+			forkHeight:   1,
+			forkHeights:  []uint64{1, 2},
```
Thegaram, line 122: Missing test case: Same fork number multiple times. E.g. in our case: `{0, 0, 0, 100}`.

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,47 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+	hardForkNumber, err := bp.getHardForkNumberByName(getTaskParameter.HardForkName)
+	if err != nil {
+		log.Error("hard fork get empty batch because of the hard fork name don't exist", "fork name", getTaskParameter.HardForkName)
+		return nil, err
+	}
+
+	// if the hard fork number set, rollup relayer must generate the chunk from hard fork number,
+	// so the hard fork chunk's start_block_number must be ForkBlockNumber
+	var startChunkIndex uint64 = 0
+	var endChunkIndex uint64 = math.MaxInt64
```
Thegaram, line 81:
```suggestion
	var endChunkIndex uint64 = math.MaxUint64
```

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,47 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+	fromBlockNum, toBlockNum := forks.BlockRange(hardForkNumber, bp.forkHeights)
+	if fromBlockNum != 0 {
+		startChunk, chunkErr := bp.chunkOrm.GetChunkByStartBlockNumber(ctx, fromBlockNum)
+		if chunkErr != nil {
+			log.Error("failed to get fork start chunk failure", "forkName", getTaskParameter.HardForkName, "fromBlockNumber", fromBlockNum, "err", chunkErr)
```
Thegaram:
```suggestion
			log.Error("failed to get fork start chunk index", "forkName", getTaskParameter.HardForkName, "fromBlockNumber", fromBlockNum, "err", chunkErr)
```

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,47 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+		}
+		if startChunk == nil {
+			return nil, nil
+		}
+		startChunkIndex = startChunk.Index
+	}
+	if toBlockNum != math.MaxUint64 {
+		toChunk, chunkErr := bp.chunkOrm.GetChunkByStartBlockNumber(ctx, toBlockNum)
+		if err != nil {
+			log.Error("failed to get fork start chunk failure", "forkName", getTaskParameter.HardForkName, "fromBlockNumber", fromBlockNum, "err", chunkErr)
```
Thegaram:
```suggestion
			log.Error("failed to get fork end chunk index", "forkName", getTaskParameter.HardForkName, "toBlockNumber", toBlockNum, "err", chunkErr)
```

coordinator/internal/logic/provertask/prover_task.go
```diff
@@ -63,6 +72,14 @@ func (b *BaseProverTask) checkParameter(ctx *gin.Context, getTaskParameter *coor
 		return nil, fmt.Errorf("incompatible prover version. please upgrade your prover, minimum allowed version: %s, actual version: %s", b.cfg.ProverManager.MinProverVersion, proverVersion.(string))
 	}
 
+	// if the prover than v4.3.41-144c7ed-8f17df8-6e1a5cd and param forkBlockNumber is empty, represents
```
Thegaram: Why this version?

coordinator/internal/logic/provertask/prover_task.go
```diff
@@ -63,6 +72,14 @@ func (b *BaseProverTask) checkParameter(ctx *gin.Context, getTaskParameter *coor
...
+	// the prover use the wrong config. don't assign task to it.
+	if version.CheckScrollRepoVersion(proverVersion.(string), "v4.3.41-144c7ed-8f17df8-6e1a5cd") {
+		if getTaskParameter.HardForkName == "" {
+			return nil, fmt.Errorf("prover version large than v4.3.41-144c7ed-8f17df8-6e1a5cd, but fork_number is empty, actual version: %s", proverVersion.(string))
```
Thegaram:
```suggestion
			return nil, fmt.Errorf("missing fork_number in get_task query, prover version: %s", proverVersion.(string))
```

coordinator/internal/logic/submitproof/proof_receiver.go
```diff
@@ -203,17 +203,17 @@ func (m *ProofReceiverLogic) HandleZkProof(ctx *gin.Context, proofMsg *message.P
 }
 
 func (m *ProofReceiverLogic) checkAreAllChunkProofsReady(ctx context.Context, chunkHash string) error {
-	batchHash, err := m.chunkOrm.GetChunkBatchHash(ctx, chunkHash)
+	batchHash, err := m.chunkOrm.GetChunkByHash(ctx, chunkHash)
```
Thegaram:
```suggestion
	chunk, err := m.chunkOrm.GetChunkByHash(ctx, chunkHash)
```

coordinator/internal/types/get_task.go
```diff
@@ -2,7 +2,8 @@ package types
 
 // GetTaskParameter for ProverTasks request parameter
 type GetTaskParameter struct {
-	ProverHeight int    `form:"prover_height" json:"prover_height"`
+	HardForkName string `form:"hard_fork_name" json:"hard_fork_name"`
```
Thegaram, line 5: Umarshal will succeed if this is missing, right?

Thegaram (commented):
Looks good

common/forks/forks_test.go
omerfirmak: Assuming `forkHeights` is the result of `CollectSortedForkHeights`, `forkHeights` won't have duplicates.

coordinator/internal/config/config.go
```diff
@@ -70,3 +73,17 @@ func NewConfig(file string) (*Config, error) {
 
 	return cfg, nil
 }
+
+// ReadGenesis parses and returns the genesis file at the given path
+func ReadGenesis(genesisPath string) (*core.Genesis, error) {
```
omerfirmak: Any chance we can get away with not duplicating this? We can try to move the definition in `rollup/internal/config/config.go` to someplace more common.

coordinator/internal/types/get_task.go
omerfirmak: Correct, Go json package treats all fields as optional.

coordinator/internal/logic/provertask/prover_task.go
```diff
@@ -91,3 +108,18 @@ func (b *BaseProverTask) checkParameter(ctx *gin.Context, getTaskParameter *coor
 	}
 	return &ptc, nil
 }
+
+func (b *BaseProverTask) getHardForkNumberByName(forkName string) (uint64, error) {
```
omerfirmak, line 104: For future reference, I believe we should make sure that multiple hard forks are not scheduled on the same non-zero block height. Because there is no way we can figure out if a prover supports all of those hard forks or just some.

coordinator/internal/logic/provertask/prover_task.go
omerfirmak: This change is not really related to hard forks, right? If not, would it make sense to move it to its own PR?

coordinator/internal/logic/provertask/chunk_prover_task.go
```diff
@@ -63,13 +66,24 @@ func (cp *ChunkProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
 	}
 
+	hardForkNumber, err := cp.getHardForkNumberByName(getTaskParameter.HardForkName)
+	if err != nil {
+		log.Error("hard fork get empty chunk because of the hard fork name don't exist", "fork name", getTaskParameter.HardForkName)
+		return nil, err
+	}
+
+	fromBlockNum, toBlockNum := forks.BlockRange(hardForkNumber, cp.forkHeights)
+	if toBlockNum > getTaskParameter.ProverHeight {
```
omerfirmak, line 76: what if fromBlockNum > ProverHeight?

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,47 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+			return nil, ErrCoordinatorInternalFailure
+		}
+		if toChunk != nil {
+			// if toChunk is empty, so the startChunk is the last chunk, don't need change the endChunkIndex of math.MaxInt64
+			endChunkIndex = toChunk.Index
```
omerfirmak: > so the startChunk is the last chunk

this is not necessarily true, there might be more chunks after startChunk. IIUC, toChunk being nil only indicates that we haven't yet reached the fork boundary. This doesn't change anything in the code tho, just a remark on the comment here.

common/forks/forks_test.go
Thegaram: Makes sense. But then this won't work in the general case, e.g. if the prover has version `london`, the correct range should be `[0, 0)` but it will return `[0, 100)`.

But that's not an issue in practice. In fact, I think we should remove all irrelevant forks from `CollectSortedForkHeights`, so for now we should only keep `shanghai` and `banach`.

coordinator/internal/logic/provertask/prover_task.go
Thegaram: If forks `A` and `B` happen on the same block `N`, I think the correct behaviour is this:
- If prover provides version `A`, then the block range will be empty (`height >= N && height < N`). This is basically a prover configuration error.
- If prover provides version `B`, then we can assign post-fork blocks to it (`height >= N`).

That said, I think we can keep only relevant forks in the monorepo fork config and remove all other forks (see my comment above).

coordinator/internal/logic/provertask/chunk_prover_task.go
Thegaram: I also considered this. I think it's implicitly considered here: we'll select `height >= forkBlockNum && `height < proverHeight` which will be an empty range.

coordinator/internal/logic/provertask/batch_prover_task.go
Thegaram: I also don't get the "so the startChunk is the last chunk" part.

I think what's meant here is: "if toChunk is empty, we don't do anything, so `endChunkIndex` will maintain its initial value of `math.MaxInt64`."

coordinator/internal/logic/provertask/chunk_prover_task.go
omerfirmak: Makes sense, might be beneficial to put an explicit check here as well. Up to @georgehao

common/forks/forks_test.go
georgehao: yes, I assume this won't happen. Because the sql will like this `select .... where start_block_number >= 0 and end_blcok_number <0`.

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: if use math.MaxUint64, the sql will failure. Postgres support the max integer is math.MaxInt64.  `在PostgreSQL 中，整数（integer）类型的上限是2,147,483,647，下限是-2,147,483,648.`

coordinator/internal/logic/provertask/prover_task.go
georgehao: ok, just a additional check, not very important, can remove it.

coordinator/internal/config/config.go
georgehao: sure

coordinator/internal/logic/provertask/prover_task.go
georgehao: fixed [49ae3c0](https://github.com/scroll-tech/scroll/pull/1182/commits/49ae3c0ceed33ce259d1adeef9980f912ae5cdfe)

coordinator/internal/logic/provertask/prover_task.go
georgehao: fixed [49ae3c0](https://github.com/scroll-tech/scroll/pull/1182/commits/49ae3c0ceed33ce259d1adeef9980f912ae5cdfe)

coordinator/internal/logic/submitproof/proof_receiver.go
georgehao: fixed [93c41f9](https://github.com/scroll-tech/scroll/pull/1182/commits/93c41f9a80b6c61409360f3f47f54827f8c04835)

common/forks/forks_test.go
Thegaram: No, what I mean, is this edge case:

- Example: `londonBlock = 0, shanghaiBlock = 0`.
- Prover provides `forkName = "london"`.
- `BlockRange` should return `0, 0` but it will incorrectly return `0, maxUint`.

coordinator/internal/logic/provertask/batch_prover_task.go
Thegaram: Okay. But in `GetChunkByStartBlockNumber` you use `math.MaxUint64` (returned from `BlockRange`), will that work?

coordinator/internal/config/config.go
georgehao: fixed [cbb966c](https://github.com/scroll-tech/scroll/pull/1182/commits/cbb966c4ba1bc83dffdfba5674ff5452af500aa3)

coordinator/internal/logic/provertask/prover_task.go
georgehao: why A and B hard fork can happen at the block?

coordinator/internal/logic/provertask/chunk_prover_task.go
georgehao: > we'll select height >= forkBlockNum && height < proverHeight` which will be an empty range.

yeah. the sql will check it. I think don't need other check.

coordinator/internal/logic/provertask/prover_task.go
Thegaram: An example is all those forks activated at `block: 0`.

But this probably won't happen for forks with height `> 0`.

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: for examples:

hard fork number : [ 1, 2]

prover hardForkNUmber : 1

block --> block number 1
chunk ---> chunk index 1 | start_block_number 1 | end_block_number 1
batch --- > batch index 1 | start_chunk_index 1 | end_chunk_index 1

after : `fromBlockNum, toBlockNum := forks.BlockRange(hardForkNumber, bp.forkHeights)`
fromBlockNum := 1
endBlockNum := 2

Then `bp.chunkOrm.GetChunkByStartBlockNumber(ctx, toBlockNum)` to get the chunk, it will be null. 

so, the comment is for this case.

> IIUC, toChunk being nil only indicates that we haven't yet reached the fork boundary

@omerfirmak's comment is more suitable.

georgehao: I'll change the comment.

georgehao: fixed [92f5abe](https://github.com/scroll-tech/scroll/pull/1182/commits/92f5abefb95b917f725bd914e2d015346accaab3)

common/forks/forks_test.go
georgehao: Yeah, currently, it return `0, maxUnit`. Let me check whether need update it

georgehao:
```
func BlockRange(forkHeight uint64, forkHeights []uint64) (from, to uint64) {
	length := len(forkHeights)
	if length == 0 {
		return 0, 0 
	}

	to = math.MaxUint64

	for i, height := range forkHeights {
		if forkHeight == height {
			from = height
			if i < length-1 {
				to = forkHeights[i+1]
			} else {
				to = math.MaxUint64
			}
			return
		} else if forkHeight < height {
			to = height
			if i != 0 {
				from = forkHeights[i-1]
			} else {
				from = 0
			}
			return
		}
		from = height
	}

	if forkHeight > forkHeights[length-1] {
		from = forkHeights[length-1]
	}
	return
}
```

Can run success all the case, but it's too complicated. And forkHeights don't have duplicated number. So How about  check forkHeights can't have duplicated number?

georgehao: add the boundary check [7f6f16f](https://github.com/scroll-tech/scroll/pull/1182/commits/7f6f16fc573694e617b5f773504f78a9da3a8b52)

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: No problem. I think scroll block number will don't exceed math.MaxInt64

coordinator/internal/logic/provertask/prover_task.go
georgehao: got it

common/forks/forks_test.go
Thegaram: We currently do have duplicates, so this check will always break.

It's okay if we don't handle the case described in the `0, 0` example above. Let's just be aware of it.

coordinator/internal/logic/provertask/batch_prover_task.go
Thegaram:
```go
toBlockNum := math.MaxUint64
// ...
toChunk, chunkErr := bp.chunkOrm.GetChunkByStartBlockNumber(ctx, toBlockNum)
```

Didn't you say above that the SQL query will fail with this?

common/forks/forks_test.go
georgehao: ok, rollback it

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: yes

georgehao: as slack discuss, Update the BlockRange's toBlock from `math.MaxUint64` to `math.MaxInt64`

common/forks/forks.go
```diff
@@ -1,37 +1,54 @@
...
 	} {
-		if fork == nil {
+		if fork.block == nil {
+			continue
+		}
+
+		height := fork.block.Uint64()
+		if height == 0 {
+			continue
+		}
```
Thegaram: If you do this, `getHardForkNumberByName` will return an error.
```suggestion
```

georgehao: this is the original logic 

https://github.com/scroll-tech/scroll/blob/develop/common/forks/forks.go#L32-L34 

maybe @omerfirmak don't think the hard can from 0, or other consideration？

Thegaram: This would mostly happen during testing, where we might set `banachBlock = 0`.

This is not relevant for the "respect fork boundary" task.

But if the prover provides `forkName = "banach"`, we should be able to assign tasks to it, even if `banachBlock = 0`.

omerfirmak: Changing this now might require revisiting the previous changes we made as well.
 
My thinking was; if a fork is enabled on genesis it is not really a fork but the initial configuration of the chain. So, every actor taking part in the network is expected to support at least the initial configuration. I would say `forkName` should only be used for forks with non-zero heights.

Thegaram: Yes, but the problem is that this might differ from network to network. Not just on a test env, if we launch a new instance of Scroll in the future, banach will probably be set at `0`.

How about we store all non-zero forks in `forkNameHeightMap` (current impl), plus we also store the last genesis fork (e.g. `shanghai` on Scroll mainnet) in this mapping?

omerfirmak: > How about we store all non-zero forks in forkNameHeightMap (current impl), plus we also store the last genesis fork (e.g. shanghai on Scroll mainnet) in this mapping?

Sounds good. We probably need to remove `if _, ok := forkHeightsMap[height]; ok {` check as well

> Yes, but the problem is that this might differ from network to network. Not just on a test env, if we launch a new instance of Scroll in the future, banach will probably be set at 0.

I see, but unless we want to share provers between Scroll instances I don't think there will be an issue.

georgehao: as discuss, add the genesis hard fork. [2fafb34](https://github.com/scroll-tech/scroll/pull/1182/commits/2fafb34a7301c382cf9870109a755056f0c08402)

common/forks/forks_test.go
```diff
@@ -17,16 +17,19 @@ func TestCollectSortedForkBlocks(t *testing.T) {
...
 		3: true,
 		4: true,
+		0: true,
 	}, m)
 	require.Equal(t, map[string]uint64{
-		"eip155": 4,
-		"eip158": 3,
+		"eip155":         4,
+		"eip158":         3,
+		"constantinople": 0,
```
omerfirmak, line 32: is it intentional that byzantium is not present in this map?

georgehao: for before, it's intentional

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -61,13 +69,48 @@ func (bp *BatchProverTask) Assign(ctx *gin.Context, getTaskParameter *coordinato
...
+		if chunkErr != nil {
+			log.Error("failed to get fork start chunk index", "forkName", getTaskParameter.HardForkName, "fromBlockNumber", fromBlockNum, "err", chunkErr)
+			return nil, ErrCoordinatorInternalFailure
+		}
+		if startChunk == nil {
+			return nil, nil
+		}
+		startChunkIndex = startChunk.Index
+	}
+	if toBlockNum != math.MaxUint64 {
```
Thegaram:
```suggestion
	if toBlockNum != math.MaxInt64 {
```

Thegaram (commented):
Will approve after this is fixed.

common/forks/forks_test.go
Thegaram: Suggestion: https://github.com/scroll-tech/scroll/pull/1205

georgehao: merged

Thegaram (dismissed):

tests/integration-test/integration_test.go
```diff
@@ -112,7 +112,7 @@ func TestCoordinatorProverInteraction(t *testing.T) {
 	t.Log(version.Version)
 
 	base.Timestamp = time.Now().Nanosecond()
-	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json")
+	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json", "./genesis.json")
```
colinlyguo, line 115:
```suggestion
	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json", "../../common/docker/l2geth/genesis.json")
```

tests/integration-test/integration_test.go
```diff
@@ -147,7 +147,7 @@ func TestProverReLogin(t *testing.T) {
 	assert.NoError(t, migrate.ResetDB(base.DBClient(t)))
 
 	base.Timestamp = time.Now().Nanosecond()
-	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json")
+	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json", "./genesis.json")
```
colinlyguo, line 150:
```suggestion
	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json", ".../../common/docker/l2geth/genesis.json")
```

colinlyguo, line 150:
```suggestion
	coordinatorApp := capp.NewCoordinatorApp(base, "../../coordinator/conf/config.json", "../../common/docker/l2geth/genesis.json")
```

colinlyguo: can use a single l2geth's genesis file, preventing maintaining multiple copies.

coordinator/internal/orm/chunk.go
```diff
@@ -196,18 +198,33 @@ func (o *Chunk) CheckIfBatchChunkProofsAreReady(ctx context.Context, batchHash s
...
+func (o *Chunk) GetChunkByHash(ctx context.Context, chunkHash string) (*Chunk, error) {
 	db := o.db.WithContext(ctx)
 	db = db.Model(&Chunk{})
 	db = db.Where("hash = ?", chunkHash)
-	db = db.Select("batch_hash")
 
 	var chunk Chunk
 	if err := db.First(&chunk).Error; err != nil {
-		return "", fmt.Errorf("Chunk.GetChunkBatchHash error: %w, chunk hash: %v", err, chunkHash)
+		return nil, fmt.Errorf("Chunk.GetChunkBatchHash error: %w, chunk hash: %v", err, chunkHash)
```
colinlyguo:
```suggestion
		return nil, fmt.Errorf("Chunk.GetChunkByHash error: %w, chunk hash: %v", err, chunkHash)
```

common/forks/forks_test.go
```diff
@@ -64,3 +72,71 @@ func TestBlocksUntilFork(t *testing.T) {
...
+		},
+		{
+			name:         "To200-1",
+			forkHeight:   100,
+			forkHeights:  []uint64{100, 200},
+			expectedFrom: 100,
+			expectedTo:   200,
+		},
+		{
+			name:         "to2",
```
colinlyguo:
```suggestion
			name:         "To2",
```

common/forks/forks_test.go
```diff
@@ -64,3 +72,71 @@ func TestBlocksUntilFork(t *testing.T) {
...
+			forkHeight:   1,
+			forkHeights:  []uint64{1, 2},
+			expectedFrom: 1,
+			expectedTo:   2,
+		},
+		{
+			name:         "to0",
```
colinlyguo:
```suggestion
			name:         "ToInfinite-1",
```

colinlyguo: same naming as the `To200-1` case.

common/forks/forks.go
```diff
@@ -54,3 +72,19 @@ func BlocksUntilFork(blockHeight uint64, forkHeights []uint64) uint64 {
 	}
 	return 0
 }
+
+// BlockRange returns the block range of the hard fork
+func BlockRange(forkHeight uint64, forkHeights []uint64) (from, to uint64) {
```
colinlyguo:
```suggestion
func BlockRange(currentForkHeight uint64, forkHeights []uint64) (from, to uint64) {
```

common/forks/forks.go
```diff
@@ -54,3 +72,19 @@ func BlocksUntilFork(blockHeight uint64, forkHeights []uint64) uint64 {
...
+	to = math.MaxInt64
+	for i, height := range forkHeights {
+		if forkHeight < height {
```
colinlyguo:
```suggestion
		if currentForkHeight < height {
```

common/forks/forks.go
```diff
@@ -54,3 +72,19 @@ func BlocksUntilFork(blockHeight uint64, forkHeights []uint64) uint64 {
...
+			to = height
+			if i != 0 {
+				from = forkHeights[i-1]
+			}
```
colinlyguo:
```suggestion
```

colinlyguo: no need to set this again because `from` is set in each iteration. see below:
```
from = height
```

coordinator/cmd/api/app/mock_app.go
```diff
@@ -95,14 +103,31 @@ func (c *CoordinatorApp) MockConfig(store bool) error {
 	cfg.Auth.LoginExpireDurationSec = 1
 	c.Config = cfg
 
+	genesis, err := utils.ReadGenesis(c.chainConfigOriginFile)
+	if err != nil {
+		return err
+	}
+	chainConf := genesis.Config
+	c.ChainConfig = chainConf
```
colinlyguo:
```suggestion
	c.ChainConfig = genesis.Config
```

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -27,16 +30,21 @@ type BatchProverTask struct {
...
+func NewBatchProverTask(cfg *config.Config, chainCfg *params.ChainConfig, db *gorm.DB, vk string, reg prometheus.Registerer) *BatchProverTask {
+	forkHeights, _, nameForkMap := forks.CollectSortedForkHeights(chainCfg)
+	log.Info("new batch prover task", "forkHeights", forkHeights)
+
 	bp := &BatchProverTask{
 		BaseProverTask: BaseProverTask{
 			vk:                 vk,
 			db:                 db,
 			cfg:                cfg,
+			nameForkMap:        nameForkMap,
```
colinlyguo, line 46: Would `nameToForkMap` here be more readable?

coordinator/internal/logic/provertask/chunk_prover_task.go
```diff
@@ -21,24 +23,25 @@ import (
...
-func NewChunkProverTask(cfg *config.Config, db *gorm.DB, vk string, reg prometheus.Registerer) *ChunkProverTask {
+func NewChunkProverTask(cfg *config.Config, chainCfg *params.ChainConfig, db *gorm.DB, vk string, reg prometheus.Registerer) *ChunkProverTask {
+	forkHeights, _, nameForkMap := forks.CollectSortedForkHeights(chainCfg)
+	log.Info("new chunk prover task", "forkHeights", forkHeights)
 	cp := &ChunkProverTask{
 		BaseProverTask: BaseProverTask{
 			vk:                 vk,
 			db:                 db,
 			cfg:                cfg,
+			nameForkMap:        nameForkMap,
```
colinlyguo, line 43: same suggestion.

coordinator/internal/logic/provertask/prover_task.go
```diff
@@ -13,6 +13,12 @@ import (
 	coordinatorType "scroll-tech/coordinator/internal/types"
 )
 
+// ErrCoordinatorInternalFailure coordinator internal db failure
+var ErrCoordinatorInternalFailure = fmt.Errorf("coordinator internal error")
+
+// ErrHardForkName client request with the wrong hard fork name
```
colinlyguo:
```suggestion
// ErrHardForkName indicates client request with the wrong hard fork name
```

coordinator/test/api_test.go
```diff
@@ -70,7 +84,7 @@ func randomURL() string {
 	return fmt.Sprintf("localhost:%d", 10000+2000+id.Int64())
 }
 
-func setupCoordinator(t *testing.T, proversPerSession uint8, coordinatorURL string) (*cron.Collector, *http.Server) {
+func setupCoordinator(t *testing.T, proversPerSession uint8, coordinatorURL string, forkNameNumber map[string]int64) (*cron.Collector, *http.Server) {
```
colinlyguo: `nameToForkMap` naming as other places.

coordinator/internal/logic/provertask/chunk_prover_task.go
```diff
@@ -21,24 +23,25 @@ import (
...
 	chunkAttemptsExceedTotal prometheus.Counter
-	chunkTaskGetTaskTotal    prometheus.Counter
+	chunkTaskGetTaskTotal    *prometheus.CounterVec
 }
 
 // NewChunkProverTask new a chunk prover task
-func NewChunkProverTask(cfg *config.Config, db *gorm.DB, vk string, reg prometheus.Registerer) *ChunkProverTask {
+func NewChunkProverTask(cfg *config.Config, chainCfg *params.ChainConfig, db *gorm.DB, vk string, reg prometheus.Registerer) *ChunkProverTask {
+	forkHeights, _, nameForkMap := forks.CollectSortedForkHeights(chainCfg)
+	log.Info("new chunk prover task", "forkHeights", forkHeights)
```
colinlyguo: can change it to `nameToForkMap` for better readability.

colinlyguo (dismissed):
The core logic LGTM. Just some typo fixes and small refactor suggestions.

tests/integration-test/integration_test.go
georgehao: Can't use the same one, because need set `homesteadBlock` and `istanbulBlock` hard fork number

colinlyguo: Got it. If wanting to keep only one copy, some settings won't break the l2geth's launch, thus are compatible with other unit tests, e.g.,
```
    "archimedesBlock": 1,
    "shanghaiBlock": 2,
```
(btw, seems that archimedesBlock is not used).

georgehao: image

Because the integrate test's block number is 1
hard fork number : [ 1, 2]

prover hardForkNUmber : 1

block --> block number 1
chunk ---> chunk index 1 | start_block_number 1 | end_block_number 1
batch --- > batch index 1 | start_chunk_index 1 | end_chunk_index 1

after : fromBlockNum, toBlockNum := forks.BlockRange(hardForkNumber, bp.forkHeights)
fromBlockNum := 1
endBlockNum := 2

Then bp.chunkOrm.GetChunkByStartBlockNumber(ctx, toBlockNum) to get the the chunk index 1

actually the `istanbulBlock` is not used. But it's not matter.

If don't set `istanbulBlock`, the from fromBlockNum=1 and endBlockNum will be math.MaxInt64

coordinator/internal/logic/provertask/batch_prover_task.go
georgehao: No, it assign value to the BaseProverTask
image

colinlyguo (dismissed):

colinlyguo:
btw, the patch version of `major.minor.patch` of scroll monorepo seems very large now.

Thegaram (dismissed):

colinlyguo (approved):

Thegaram (approved):

colinlyguo:
tweaked the PR's name, and can further consider it.

PR #1183, opened 2024-03-08
feat(sender): support blob transaction

colinlyguo:

Refactor sender to support sending and resubmitting blob transactions.

- [x] Supporting initializing a post-merge private network in unit test to test blob transactions.
- [x] Adding unit tests.

### PR title

- [x] feat: A new feature
- [x] refactor: A code change that doesn't fix a bug, or add a feature, or improves performance

### Deployment tag versioning

- [x] Yes

### Breaking change label


common/docker-compose/l1/execution/genesis.json
```diff
@@ -0,0 +1,127 @@
...
+	},
+	"nonce": "0x0",
+	"timestamp": "0x65f19c6f",
+	"extraData": "0x0000000000000000000000000000000000000000000000000000000000000000123463a4b065722e99115d6c222f267d9cabb5240000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+	"gasLimit": "0x1c9c380",
+	"difficulty": "0x1",
+	"mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+	"coinbase": "0x0000000000000000000000000000000000000000",
+	"alloc": {
+		"03a1bba60b5aa37094cf16123add674c01589488": {
```
Thegaram: Are these accounts used in tests?

common/docker-compose/l1/execution/sk.json
Thegaram, line 1: What's this file used for?

rollup/internal/controller/sender/estimategas.go
```diff
@@ -33,15 +34,19 @@ func (s *Sender) estimateLegacyGas(to *common.Address, value *big.Int, data []by
...
 	gasTipCap, err := s.client.SuggestGasTipCap(s.ctx)
 	if err != nil {
 		log.Error("estimateDynamicGas SuggestGasTipCap failure", "error", err)
 		return nil, err
 	}
 
-	gasFeeCap := new(big.Int).Add(gasTipCap, new(big.Int).Mul(new(big.Int).SetUint64(baseFee), big.NewInt(2)))
-	gasLimit, accessList, err := s.estimateGasLimit(to, data, nil, gasTipCap, gasFeeCap, value, true)
+	// gasFeeCap = baseFee * 2 + gasTipCap
+	gasFeeCap := new(big.Int).SetUint64(baseFee)
```
Thegaram:
```go
gasFeeCap := getGasFeeCap(new(big.Int).SetUint64(baseFee), gasTipCap)
```

rollup/internal/controller/sender/sender.go
```diff
@@ -373,13 +396,63 @@ func (s *Sender) resubmitTransaction(tx *gethTypes.Transaction, baseFee uint64)
...
+
+		// adjust for rising blobbasefee
+		currentBlobGasFeeCap := getBlobGasFeeCap(new(big.Int).SetUint64(blobBaseFee))
+		if blobGasFeeCap.Cmp(currentBlobGasFeeCap) < 0 {
+			blobGasFeeCap = currentBlobGasFeeCap
+		}
+
+		// but don't exceed maxGasPrice
+		if blobGasFeeCap.Cmp(maxGasPrice) > 0 {
+			blobGasFeeCap = maxGasPrice
```
Thegaram: Use the same `maxGasPrice` for execution and blob gas?

rollup/internal/controller/sender/sender_test.go
```diff
@@ -548,3 +566,71 @@ func testCheckPendingTransactionTxMultipleTimesWithOnlyOneTxPending(t *testing.T
...
+	// exit by ctx.Done()
+	cfgCopy2 := *cfg.L1Config.RelayerConfig.SenderConfig
+	cfgCopy2.TxType = BlobTxType
+	subCtx, cancel := context.WithCancel(context.Background())
+	_, err = NewSender(subCtx, &cfgCopy2, blobSenderPrivateKey, "test", "test", types.SenderTypeUnknown, db, nil)
+	assert.NoError(t, err)
+	cancel()
+}
+
+func testSendAndRetrieveBlobTransaction(t *testing.T) {
```
Thegaram: Is it possible to test the blob gas bump logic somehow?

Thegaram (dismissed):

rollup/internal/controller/sender/sender.go
colinlyguo: I reuse the config. wdyt? separating them or using the same one?

rollup/internal/controller/sender/sender_test.go
colinlyguo: added in [c08caa0](https://github.com/scroll-tech/scroll/pull/1183/commits/c08caa06a2463a5cffddb822d33721cf9df744c0)

common/docker-compose/l1/execution/sk.json
colinlyguo: it's not used. removed.

common/docker-compose/l1/execution/genesis.json
colinlyguo: removed the unused accounts and contract.

rollup/internal/controller/sender/estimategas.go
colinlyguo: fixed in [1f8c930](https://github.com/scroll-tech/scroll/pull/1183/commits/1f8c9308faa91ff64a6bca162732c01164aa5d28)

rollup/internal/controller/sender/sender.go
Thegaram: Let's split this into 2 configs

bridge-history-api/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/go-redis/redis/v8 v8.11.5
 	github.com/pressly/goose/v3 v3.16.0
 	github.com/prometheus/client_golang v1.14.0
-	github.com/scroll-tech/go-ethereum v1.10.14-0.20240311135752-ccec84ce63c8
+	github.com/scroll-tech/go-ethereum v1.10.14-0.20240313115705-5bd1affaeb50
```
Thegaram: [4553f5f](https://github.com/scroll-tech/go-ethereum/commit/4553f5f2693542c20b5c68080ae5bacad12e4cf2)

otara81:
Wondwrfull

bridge-history-api/go.mod
colinlyguo: updated.

Makefile
```diff
@@ -43,14 +43,15 @@ fmt: ## format the code
 	goimports -local $(PWD)/tests/integration-test/ -w .
 
 dev_docker: ## build docker images for development/testing usages
-	docker build -t scroll_l1geth ./common/docker/l1geth/
-	docker build -t scroll_l2geth ./common/docker/l2geth/
+	docker pull postgres
+	docker build -t scroll_l1geth ./common/docker/l1geth/ --platform linux/amd64
+	docker build -t scroll_l2geth ./common/docker/l2geth/ --platform linux/amd64
```
georgehao: I also encounter this issue, but not sure whether need add this. if someone run this on x86_64 maybe have issue

colinlyguo: Got it. I change this because CI and Mac can both use these commands. forget to consider a broader use case.

common/docker-compose/l1/docker-compose.go
```diff
@@ -0,0 +1,249 @@
+package dockercompose
+
+import (
+	"crypto/rand"
+	"fmt"
+	"math/big"
+	"os"
+	"os/exec"
+	"path/filepath"
```
georgehao, line 9: It's better to use testcontainer

colinlyguo: make sense. will change it.

Makefile
colinlyguo: but I think we only added images of `linux/amd64`? thus can keep this. cc @Thegaram for more context about image building.

common/docker-compose/l1/docker-compose.go
colinlyguo: changed.

Makefile
colinlyguo: removed.

rollup/internal/controller/sender/sender.go
colinlyguo: added in [4be6a10](https://github.com/scroll-tech/scroll/pull/1183/commits/4be6a105176a7728e7ab4a10682430087974af19)

rollup/Makefile
```diff
@@ -4,8 +4,8 @@ IMAGE_VERSION=latest
 REPO_ROOT_DIR=./..
 
 mock_abi:
-	cd .. && go run github.com/scroll-tech/go-ethereum/cmd/abigen --sol ./rollup/mock_bridge/MockBridgeL1.sol --pkg mock_bridge --out ./rollup/mock_bridge/MockBridgeL1.go
-	cd .. && go run github.com/scroll-tech/go-ethereum/cmd/abigen --sol ./rollup/mock_bridge/MockBridgeL2.sol --pkg mock_bridge --out ./rollup/mock_bridge/MockBridgeL2.go
+	cd .. && solc --evm-version cancun --bin --abi --optimize --overwrite -o ./build/bin ./rollup/mock_bridge/MockBridge.sol
```
Thegaram, line 7: Note: For L2 contracts we don't support `cancun`. I don't think the `solc` compiler uses any of the new opcodes by default, though. Might become an issue with future versions, something to keep an eye on.

rollup/internal/controller/sender/sender.go
```diff
@@ -418,7 +492,7 @@ func (s *Sender) checkPendingTransaction() {
 		}
 
 		receipt, err := s.client.TransactionReceipt(s.ctx, tx.Hash())
-		if (err == nil) && (receipt != nil) { // tx confirmed.
+		if err == nil { // tx confirmed.
```
Thegaram, line 495: What's the rationale for this change?

rollup/internal/controller/sender/sender.go
```diff
@@ -373,13 +397,63 @@ func (s *Sender) resubmitTransaction(tx *gethTypes.Transaction, baseFee uint64)
...
+	default:
+		return nil, fmt.Errorf("unsupported transaction type: %s", s.config.TxType)
 	}
 
 	log.Info("Transaction gas adjustment details", "service", s.service, "name", s.name, "txInfo", txInfo)
 
 	nonce := tx.Nonce()
 	s.metrics.resubmitTransactionTotal.WithLabelValues(s.service, s.name).Inc()
-	tx, err := s.createAndSendTx(&feeData, tx.To(), tx.Value(), tx.Data(), &nonce)
+	tx, err := s.createAndSendTx(&feeData, tx.To(), tx.Data(), tx.BlobTxSidecar(), &nonce)
```
Thegaram, line 456: Will this work in the graceful restart path (inside `checkPendingTransaction`)? Or do we need to add the blob data to `PendingTransaction`?

rollup/mock_bridge/MockBridge.sol
```diff
@@ -312,4 +312,35 @@ contract MockBridgeL1 {
       mstore(memPtr, dataHash)
     }
   }
+
+  address private constant POINT_EVALUATION_PRECOMPILE_ADDRESS = 0x000000000000000000000000000000000000000A;
+  uint256 private constant BLS_MODULUS = 52435875175126190479447740508185965837690552500527637822603658699938581184513;
+
+  function verifyProof(
```
Thegaram, line 319: What do we test with this method? Since this is not the same logic as the actual function that we'll add to `ScrollChain`

rollup/internal/controller/sender/sender.go
colinlyguo: Yes. Because the db only uses rlp encoding/decoding to handle transactions (the `rlp_encoding` field), other fields related to transaction info are just added for debugging.

btw, resubmission is tested in unit tests now because each unit test needs to wait until the transactions are confirmed, and many tests would trigger this logic (both explicitly added and implicitly resubmitted).

rollup/mock_bridge/MockBridge.sol
colinlyguo: It's only a smoke test with a contract call containing `blobhash` op, ensuring something wouldn't be broken in `eth_createAccessList`, `eth_estimateGas`. Previously, blob support of these JSON APIs in go SDK is still WIP, but it works fine now.

Mock interfaces for ScrollChain will be tested in `rollup-relayer`'s unit test.

rollup/Makefile
colinlyguo: I also noticed this. When solving this issue, I found that the unit tests do not use `mock L2 contract` anywhere (gas-oracle only checks db update), thus removing `mock L2 contract` temporarily (otherwise with the new `abigen` command, some Golang function names will conflict, need to restructure the mock_bridge dir's tree structure).

Will add it back when it's used.

rollup/internal/controller/sender/sender.go
colinlyguo: To simplify the if-else logic here (IMO the current version is more readable). The `TransactionReceipt` would not return this case (err == nil && receipt == nil).

now: https://github.com/scroll-tech/go-ethereum/blob/4553f5f2693542c20b5c68080ae5bacad12e4cf2/ethclient/ethclient.go#L288-L292
upstream: https://github.com/ethereum/go-ethereum/blob/65142d81dd2723b54914233a21d35b3d6ac57832/ethclient/ethclient.go#L310-L312

If the code repo does not want to rely on this (maybe?) feature of `TransactionReceipt`, other places also need this `nil` check, in `TransactionByHash`, `BlockByNumber`, etc. So I think we can keep it simple (these APIs all return `ethereum.NotFound` instead of a nil pointer and a nil error).

colinlyguo: > To simplify the if-else logic here (IMO the current version is more readable). The `TransactionReceipt` would not return this case (err == nil && receipt == nil).
> 
> now: https://github.com/scroll-tech/go-ethereum/blob/4553f5f2693542c20b5c68080ae5bacad12e4cf2/ethclient/ethclient.go#L288-L292 upstream: https://github.com/kumakichi/go-ethereum/blob/65142d81dd2723b54914233a21d35b3d6ac57832/ethclient/ethclient.go#L310-L312
> 
> if the code repo does not want to rely on this feature (maybe?) of `TransactionReceipt`, other places also need this `nil` check, in `TransactionByHash`, `BlockByNumber`, etc. So I think we can keep it simple (these APIs all return `ethereum.NotFound` instead of a nil pointer and a nil error).

Thegaram (dismissed):

Thegaram (approved):

zimpha (approved):

PR #1212, opened 2024-03-28
feat(libzkp): upgrade to `v0.10.0k` to support blob DA

lispc:
Some highlights here:

1. I changed a golang function impl to support go 1.19.  A lot of our current existed builder image use go 1.19.  So i think it is better to still support go 1.19 for a while.
2. As usuall, tested with https://github.com/scroll-tech/test_zkp/pull/6. 
3. Assets version: v0.10.0k.   Halo2-gpu version: v0.7.5
4. I also made some modifications to prover_test.go to increase flexibility and support batches with a chunk number other than 2.


common/utils/utils.go
```diff
@@ -71,10 +70,16 @@ func ReadGenesis(genesisPath string) (*core.Genesis, error) {
...
+		}
+	}()
 
 	genesis := new(core.Genesis)
 	if err := json.NewDecoder(file).Decode(genesis); err != nil {
-		return nil, errors.Join(err, file.Close())
+		return nil, err
 	}
-	return genesis, file.Close()
+	return genesis, nil
```
omerfirmak: No error will be reported to the caller if deferred close fails when returning from this line

roylou: @HAOYUatHZ can you help to address this comment and merge?
```suggestion
	return genesis, err
```

(optional) Actually you may want to have an additional variable `retErr` to distinguish from widely used local variable `err`.

PR #1249, opened 2024-04-08
feat(coordinator): support multiple batch verifier versions

lispc:

By using hard fork id

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes


coordinator/internal/logic/verifier/verifier.go
```diff
@@ -78,7 +78,8 @@ func (v *Verifier) VerifyBatchProof(proof *message.BatchProof) (bool, error) {
 	}()
 
 	log.Info("Start to verify batch proof ...")
-	verified := C.verify_batch_proof(proofStr)
+        var forkID int64 = 2 // FIXME: get forkID from block height?
+	verified := C.verify_batch_proof(proofStr, C.int64_t(forkID))
```
lispc: change: var forkName = XXX

lispc:
a lint error

```
Error: test/mock_prover.go:164:82: `(*mockProver).tryGetProverTask` - `forkName` always receives `"homestead"` (unparam)
func (r *mockProver) tryGetProverTask(t *testing.T, proofType message.ProofType, forkName string) (int, string) {
                                        ```

@georgehao

File: common/libzkp/impl/src/batch.rs [hunk 2]
```diff
@@ -148,11 +149,33 @@ pub unsafe extern "C" fn gen_batch_proof(
...
+) -> c_char {
     let proof = c_char_to_vec(proof);
     let proof = serde_json::from_slice::<BatchProof>(proof.as_slice()).unwrap();
-
-    let verified = panic_catch(|| VERIFIER.get().unwrap().verify_agg_evm_proof(proof));
+    let fork_name_str = c_char_to_str(fork_name);
+    let fork_id = match fork_name_str {
+        "" => 1,
+        "shanghai" => 1,
+        "bernoulli" => 2,
```
Thegaram: I'd suggest using `0` and `1` for consistency (as we do in `codecv0` and `codecv1`, and also in the batch header `Version` field).

common/libzkp/impl/src/evm_verifier_fork_1.bin
Thegaram, line 1: What's this verifier file? While do we use this approach for `1` but not for `2`?

common/libzkp/impl/src/batch.rs
lispc: i think we use names like "upgrade#1" "upgrade#2" "upgrade#3".  I feel being consistent with that is good?

common/libzkp/impl/src/evm_verifier_fork_1.bin
lispc: it is https://circuit-release.s3.us-west-2.amazonaws.com/release-v0.9.9/evm_verifier.bin.    For the v0.10, what really happens is also like this, use revm to interprete a EVM bytecode to verify proof. For v0.10, it is located inside the assets dir.  But for legacy 0.9 verison, i feel it it too heavy to set up a "legacy_assets" dir, and ask provers to download and manage?  So i just embed the 0.9 verifier inside codes.

common/libzkp/impl/src/batch.rs
Thegaram: Well "upgrade #1" has no significance to provers/verifiers. We're already using 0/1 everywhere (except for the names you mentioned). But we can also keep this PR as it is.

common/libzkp/impl/src/evm_verifier_fork_1.bin
Thegaram: I meant the other way around, isn't it easier for `v0.10` to just embed it, instead of reading from an asserts dir? (Not really relevant for this PR, we can discuss separately as a potential improvement.)

common/libzkp/impl/src/batch.rs
lispc: “using 0/1 everywhere”: in l2geth？then i think it may be better to be consistent

common/libzkp/impl/src/evm_verifier_fork_1.bin
lispc: Having the assets dir We can do more complex things besides just calling verifier bytecode in future if needed. So i would like to keep assets dir

common/libzkp/impl/src/batch.rs
Thegaram: Yes https://github.com/scroll-tech/go-ethereum/blob/develop/rollup/rollup_sync_service/rollup_sync_service.go#L469

lispc: resolved

coordinator/internal/logic/provertask/prover_task.go
```diff
@@ -68,12 +71,24 @@ func (b *BaseProverTask) checkParameter(ctx *gin.Context, getTaskParameter *coor
...
 
+	vk, vkExist := b.vkMap[ptc.HardForkName]
+	if !vkExist {
+		return nil, fmt.Errorf("can't get vk for hard fork:%s, vkMap:%v", ptc.HardForkName, b.vkMap)
+	}
+
 	// if the prover has a different vk
-	if getTaskParameter.VK != b.vk {
+	if getTaskParameter.VK != vk {
+		log.Error("vk.....", "prover vk", getTaskParameter.VK, "vk", vk, "hardForkName", ptc.HardForkName)
```
colinlyguo:
```suggestion
		log.Error("vk inconsistency", "prover vk", getTaskParameter.VK, "vk", vk, "hardForkName", ptc.HardForkName)
```

Thegaram (approved):

georgehao (approved):

PR #1282, opened 2024-04-19
feat(contracts): batch token bridge

zimpha:

This PR add contracts used for batch token deposits from L1 to L2.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes


contracts/src/batch-bridge/L2BatchBridgeGateway.sol
```diff
@@ -0,0 +1,239 @@
...
+        address oldL1Token = tokenMapping[l2Token];
+        tokenMapping[l2Token] = l1Token;
+
+        emit UpdateTokenMapping(l2Token, oldL1Token, l1Token);
+    }
+
+    /// @notice Withdraw distribution failed tokens.
+    /// @param token The address of token to withdraw.
+    /// @param receiver The address of token receiver.
+    function withdrawFailedAmount(address token, address receiver) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
colinlyguo, line 175: if there is an event for this, will be much easier to track the withdrawal status of failed balance (clear the sum added based on `DistributeFailed` event).

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+    function _tryFinalzeCurrentPhase(
+        address token,
+        BatchSetting memory cachedBatchSetting,
+        TokenState memory cachedTokenState
+    ) internal view {
+        if (cachedBatchSetting.maxTxPerBatch == 0) {
+            revert ErrorTokenNotSupported();
+        }
+        PhaseState memory cachedPhaseState = phases[token][cachedTokenState.currentPhaseIndex];
+        if (cachedPhaseState.numDeposits == 0) return;
```
icemelon: this line seems unnecessary

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+    /// @param numDeposits The total number of deposits in current phase.
+    /// @param hash The hash of current phase.
+    ///   Suppose there are `n` deposits in current phase with `senders` and `amounts`. The hash is computed as
+    ///   ```text
+    ///   hash[0] = concat(token, phase_index)
+    ///   hash[i] = keccak(hash[i-1], concat(senders[i], amounts[i]))
+    ///   ```
+    ///   The type of `token` and `senders` is `address`, while The type of `phase_index` and `amounts[i]` is `uint96`.
+    ///   In current way, the hash of each phase should be different.
+    struct PhaseState {
```
icemelon: Rename to `BatchState`

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+            newSetting.feeAmountPerTx > newSetting.minAmountPerTx
+        ) {
+            revert ErrorInvalidBatchSetting();
+        }
+        settings[token] = newSetting;
+    }
+
+    /// @notice Initiate the batch bridge of current pending phase.
+    /// @param token The address of the token.
+    function batchBridge(address token) external payable onlyRole(KEEPER_ROLE) {
```
icemelon:
```suggestion
    function batchDeposit(address token) external payable onlyRole(KEEPER_ROLE) {
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+    }
+
+    /// @dev Compiler will pack this into a single `bytes32`.
+    /// @param pending The total amount of token pending to bridge.
+    /// @param currentPhaseIndex The index of current phase.
+    /// @param pendingPhaseIndex The index of pending phase (next phase to batch bridge).
+    struct TokenState {
+        uint128 pending;
+        uint64 currentPhaseIndex;
+        uint64 pendingPhaseIndex;
```
icemelon:
```suggestion
        uint64 currentBatchIndex;
        uint64 pendingBatchIndex;
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+        phases[token][cachedTokenState.currentPhaseIndex] = cachedPhaseState;
+        tokens[token] = cachedTokenState;
+    }
+
+    /// @dev Internal function to finalze current phase.
+    ///      This function may change the value of `cachedTokenState`, which can be used in later operation.
+    /// @param token The address of token to finalze.
+    /// @param cachedBatchSetting The cached batch setting in memory.
+    /// @param cachedTokenState The cached token state in memory.
+    function _tryFinalzeCurrentPhase(
```
icemelon:
```suggestion
    function _tryFinalzeCurrentBatch(
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,418 @@
...
+     ***********/
+
+    /// @notice The setting for batch token bridging.
+    /// @dev Compiler will pack this into a single `bytes32`.
+    /// @param feeAmountPerTx The amount of fee charged for each deposit.
+    /// @param minAmountPerTx The minimum amount of token for each deposit.
+    /// @param maxTxPerBatch The maximum number of deposit in each batch.
+    /// @param maxDelayPerBatch The maximum number of seconds to wait in each batch.
+    /// @param safeBridgeGasLimit The safe bridge gas limit for bridging token from L1 to L2.
+    struct BatchSetting {
```
icemelon:
```suggestion
    struct BatchConfig {
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: This line is used to skip the very first batch. In the first batch, `firstDepositTimestamp` is zero. If we won't add this line, the first batch will be considered finalized.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+            cachedBatchState.hash = BatchBridgeCodec.hash(cachedBatchState.hash, node);
+        }
+
+        batches[token][cachedTokenState.currentBatchIndex] = cachedBatchState;
+        tokens[token] = cachedTokenState;
+    }
+
+    /// @dev Internal function to finalze current batch.
+    ///      This function may change the value of `cachedTokenState`, which can be used in later operation.
+    /// @param token The address of token to finalze.
```
georgehao:
```suggestion
    /// @param token The address of the finalization token
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    /// @dev Compiler will pack this into a single `bytes32`.
+    /// @param feeAmountPerTx The amount of fee charged for each deposit.
+    /// @param minAmountPerTx The minimum amount of token for each deposit.
+    /// @param maxTxPerBatch The maximum number of deposit in each batch.
+    /// @param maxDelayPerBatch The maximum number of seconds to wait in each batch.
+    /// @param safeBridgeGasLimit The safe bridge gas limit for bridging token from L1 to L2.
+    struct BatchConfig {
+        uint96 feeAmountPerTx;
+        uint96 minAmountPerTx;
+        uint16 maxTxPerBatch;
```
georgehao:
```suggestion
        uint16 maxTxsPerBatch;
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+        uint24 maxDelayPerBatch;
+        uint24 safeBridgeGasLimit;
+    }
+
+    /// @dev Compiler will pack this into two `bytes32`.
+    /// @param amount The total amount of token to deposit in current batch.
+    /// @param firstDepositTimestamp The timestamp of first deposit.
```
icemelon:
```suggestion
    /// @param startTime The timestamp of the first deposit.
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    ///   Suppose there are `n` deposits in current batch with `senders` and `amounts`. The hash is computed as
+    ///   ```text
+    ///   hash[0] = concat(token, batch_index)
+    ///   hash[i] = keccak(hash[i-1], concat(senders[i], amounts[i]))
+    ///   ```
+    ///   The type of `token` and `senders` is `address`, while The type of `batch_index` and `amounts[i]` is `uint96`.
+    ///   In current way, the hash of each batch among all tokens should be different.
+    struct BatchState {
+        uint128 amount;
+        uint64 firstDepositTimestamp;
```
icemelon:
```suggestion
        uint64 startTime;
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: `The address of the finalization token` doesn't really explain the meaning.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+import {IL1ERC20Gateway} from "../L1/gateways/IL1ERC20Gateway.sol";
+import {IL1GatewayRouter} from "../L1/gateways/IL1GatewayRouter.sol";
+import {IL1MessageQueue} from "../L1/rollup/IL1MessageQueue.sol";
+import {IL1ScrollMessenger} from "../L1/IL1ScrollMessenger.sol";
+
+import {BatchBridgeCodec} from "./BatchBridgeCodec.sol";
+import {L2BatchBridgeGateway} from "./L2BatchBridgeGateway.sol";
+
+/// @title L1BatchBridgeGateway
+contract L1BatchBridgeGateway is AccessControlEnumerableUpgradeable, ReentrancyGuardUpgradeable {
```
Thegaram, line 20: Does `L1BatchBridgeGateway` need to implement drop and replay functionalities?

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    ///
+    /// @param token The address of token.
+    /// @param amount The amount of token to deposit. We use type `uint96`, since it is enough for most of the major tokens.
+    function depositERC20(address token, uint96 amount) external {
+        if (token == address(0)) revert ErrorIncorrectMethodForETHDeposit();
+
+        // common practice to handle fee on transfer token.
+        uint256 beforeBalance = IERC20Upgradeable(token).balanceOf(address(this));
+        IERC20Upgradeable(token).safeTransferFrom(_msgSender(), address(this), amount);
+        amount = uint96(IERC20Upgradeable(token).balanceOf(address(this)) - beforeBalance);
```
Thegaram, line 227: Should we check overflow, just in case?

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    function depositETH() external payable {
+        // no safe cast check here, since no one has so much ETH yet.
+        _deposit(address(0), _msgSender(), uint96(msg.value));
+    }
+
+    /// @notice Deposit ERC20 token.
+    ///
+    /// @param token The address of token.
+    /// @param amount The amount of token to deposit. We use type `uint96`, since it is enough for most of the major tokens.
+    function depositERC20(address token, uint96 amount) external {
```
Thegaram, line 221: Would it make sense to also allow deposit with permit? (Although we didn't add this to the other gateways either.)

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    /// @param sender The address of token sender.
+    /// @param amount The amount of token to deposit.
+    function _deposit(
+        address token,
+        address sender,
+        uint96 amount
+    ) internal nonReentrant {
+        BatchConfig memory cachedBatchConfig = configs[token];
+        TokenState memory cachedTokenState = tokens[token];
+        _tryFinalzeCurrentBatch(token, cachedBatchConfig, cachedTokenState);
```
Thegaram:
```suggestion
        _tryFinalizeCurrentBatch(token, cachedBatchConfig, cachedTokenState);
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+    function _tryFinalzeCurrentBatch(
+        address token,
+        BatchConfig memory cachedBatchConfig,
+        TokenState memory cachedTokenState
+    ) internal view {
+        if (cachedBatchConfig.maxTxsPerBatch == 0) {
+            revert ErrorTokenNotSupported();
+        }
+        BatchState memory cachedBatchState = batches[token][cachedTokenState.currentBatchIndex];
+        // return if it is the very first batch
```
Thegaram:
```suggestion
        // return if it is the very first deposit in the current batch
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+            newConfig.feeAmountPerTx > newConfig.minAmountPerTx
+        ) {
+            revert ErrorInvalidBatchConfig();
+        }
+        configs[token] = newConfig;
+    }
+
+    /// @notice Initiate the batch bridge of current pending batch.
+    /// @param token The address of the token.
+    function batchDeposit(address token) external payable onlyRole(KEEPER_ROLE) {
```
Thegaram:
```suggestion
    function executeBatchDeposit(address token) external payable onlyRole(KEEPER_ROLE) {
```

Might be a clearer name

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+            IL1ERC20Gateway(gateway).depositERC20{value: depositFee}(
+                token,
+                counterpart,
+                cachedBatchState.amount,
+                cachedBatchConfig.safeBridgeGasLimit
+            );
+        }
+
+        // notify `L2BatchBridgeGateway`
+        IL1ScrollMessenger(messenger).sendMessage{value: batchBridgeFee}(
```
Thegaram, line 309: What's the rationale for having a 2nd call to the bridge (as opposed to adding this calldata to the 1st call)? Is it that this way we can use a single `finalizeBatchDeposit` function on L2 for both ETH and ERC20 tokens?

contracts/src/batch-bridge/L2BatchBridgeGateway.sol
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /************************
+     * Restricted Functions *
+     ************************/
+
+    /// @notice Update layer 2 to layer 1 token mapping.
+    /// @param l2Token The address of corresponding ERC20 token on layer 2.
+    /// @param l1Token The address of ERC20 token on layer 1.
+    function updateTokenMapping(address l2Token, address l1Token) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
Thegaram: Why don't we read the mapping from `L2GatewayRouter`?

Thegaram (dismissed):

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: The batch config has `safeBridgeGasLimit`, which should allow token bridge successfully. So no `replay` is needed. As for drop, hopefully our l2geth don't have bug to skip normal bridge transaction.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: for normal and fee on transfer token, `IERC20Upgradeable(token).balanceOf(address(this)) - beforeBalance` always fit `uint96`, since `amount` is type `uint96`.

for other malicious token, we won't add it to this contract.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: `permit` adds complexity, better not to add for now.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: Not all erc20 tokens support `depositERC20AndCall` in layer 2, so we need two `sendMessage` for erc20 tokens. And to make it consistent, I also use two `sendMessage` for ETH.

contracts/src/batch-bridge/L2BatchBridgeGateway.sol
zimpha: yes, we can. Adding mapping here is to save the bridge gas.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
georgehao: ok, `finalze` typo --> `finalize`

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
colinlyguo: also noticed this, if `one of them is skipped` but `the other one is successfully relayed`, would this happen (or will not happen in practice)?

there are 2 cases:

if `fund transfer` is skipped and `message relayed` is successful, can trigger refund to L1BatchBridge, may need to upgrade the contract, supporting refund to each deposit users in this batch.

if `message relayed` is skipped and `fund transfer` is successful, how to solve this? update: may need to upgrade the L2BatchBridge contract.

zimpha: A general assumption is that both messages won't be skipped. If so, I think our whole layer2 would have a big bug and all other bridge messages will also be skipped.

colinlyguo: oh, got it. we can construct cross-chain messages with many `ecPairing` calls to make them skipped, but messages in this pr are simple fund transfers or small payload message relays (just some metadata), they would not be skipped.

Thegaram: Why do we need `depositERC20AndCall`? I thought how it works is:
1. Tokens are bridged to L2 receiver aka `L2BatchBridgeGateway` (https://github.com/scroll-tech/scroll/blob/develop/contracts/src/L2/gateways/L2CustomERC20Gateway.sol#L100)
2. The `ERC20Gateway` calls `L2BatchBridgeGateway.onScrollGatewayCallback`.

While the scenario that Colin mentioned is unlikely to happen, it's very concerning: "fund transfer is skipped and message relayed successful" would lead to inconsistency on L1 and L2, and it would be hard to recover from. This seems worse than "other bridge messages will also be skipped".

zimpha: like said before, not all ERC20 gateway will call `L2BatchBridgeGateway.onScrollGatewayCallback`. The skip feature is designed to skip transactions that cannot be proved. In our case, it is very unlikely to happen.

Let's say, if it happens. We can upgrade the contract and do manually refund.

zimpha: also, if we worry about this. We should also worry that `L1BatchBridgeGateway.executeBatchDeposit` and `L2BatchBridgeGateway.distribute` can only be called by permissioned keeper. If we stop keeper program, funds will be locked in the contract.

colinlyguo: It makes sense.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+        _tryFinalizeCurrentBatch(token, cachedBatchConfig, cachedTokenState);
+
+        // no batch to bridge
+        if (cachedTokenState.currentBatchIndex == cachedTokenState.pendingBatchIndex) {
+            revert ErrorNoPendingBatch();
+        }
+
+        // check bridge fee
+        uint256 depositFee = IL1MessageQueue(queue).estimateCrossDomainMessageFee(cachedBatchConfig.safeBridgeGasLimit);
+        uint256 batchBridgeFee = IL1MessageQueue(queue).estimateCrossDomainMessageFee(SAFE_BATCH_BRIDGE_GAS_LIMIT);
```
icemelon, line 267: Why not combine the estimate into one single call?
```suggestion
        uint256 batchBridgeFee = IL1MessageQueue(queue).estimateCrossDomainMessageFee(SAFE_BATCH_BRIDGE_GAS_LIMIT + cachedBatchConfig.safeBridgeGasLimit);
```

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
```diff
@@ -0,0 +1,425 @@
...
+        if (msg.value < depositFee + batchBridgeFee) {
```
icemelon, line 268: Just to understand the two fees. The batch bridge fee is the gas for execution in the scroll messenger contract while the deposit fee is the gas for ETH or ERC20 transfers. Is this right?

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
icemelon: Actually why don't we make `L1BatchBridgeGateway.executeBatchDeposit` and `L2BatchBridgeGateway.distribute` permissionless?

contracts/src/batch-bridge/L2BatchBridgeGateway.sol
icemelon: We should just read the mapping from `L2GatewayRouter`. It can reduce the operational overhead when we update the ERC20 token mapping. The gas cost should be cheaper on L2, and you already call `IL1ERC20Gateway` to get the corresponding L2 token on L1.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
spilehchiha: That is how I understand it, yes.
cc @zimpha

zimpha: yeah, the `depositFee` is for bridge ETH/ERC20 from L1 to L2 and the `batchBridgeFee` is for notifying the deposit to L2.

contracts/src/batch-bridge/L1BatchBridgeGateway.sol
zimpha: permissionless is good but requires a different design (i.e. how to incentive callers) and more codes. If we go permissionless, we should definitely consider about how to handle skipped messages and the potential security problems.

It is a new feature, we should be more careful. In the future, we may change from permissioned to permissionless.

contracts/src/batch-bridge/L2BatchBridgeGateway.sol
zimpha: removed.

spilehchiha (approved):

icemelon (approved):

Thegaram (approved):

PR #1292, opened 2024-04-22
docs: add deployment addresses

TiesD:
- Added Scroll deployment addresses for L1 & L2 proxy contracts, fetched from [here](https://docs.scroll.io/en/developers/scroll-contracts/)
- Removed redundant l1geth.json & l2geth.json files


0xmountaintop:
1. l1geth.json & l2geth.json are used in the scripts in `contracts/scripts`, please do not delete
2. as mentioned in the `contracts/deployments/README.md`, the description is related to a local testnet, so it makes no sense to fill in the scroll mainnet addresses there.

PR #1303, opened 2024-04-25
feat: add deterministic, multi-layer deployment script

Thegaram:

**Problem**: The current deployment scripts are slow and error-prone. We need to predict the deployment addresses of some contracts (e.g. for preallocating funds to `L2ScrollMessenger` in L2 genesis), but currently the addresses depend on the deployer private key and the order of deployment. If one transaction fails, subsequent transactions will lead to different addresses.

**Solution**:
- Use https://github.com/Arachnid/deterministic-deployment-proxy to make contract addresses much more predictable.
- Use a single Foundry script for deployment and initialization on L1 and L2 (basically merge the previous 5 scripts).
- Move genesis generation into the Fundry script. Previously this was a complex combination of Javascript, Rush, and bash scripts.
- Provide convenient Docker images. This makes deployment much more robust (we enforce a fixed Foundry version known to be working), and fast (compilation done in the build phase, no need to recompile when running the container).
- Use a TOML configuration file.

**Caveats**:
- Some of the deployed addresses still depend on the deployer account. The reason is that some contracts take the `owner` as a constructor argument, and the initial owner is the deployer account, so changing this also changes the `create2` deterministic address. But the addresses do not depend on deployment order anymore.

### PR title

- [X] feat: A new feature

### Deployment tag versioning

- [X] No, this PR doesn't involve a new deployment, git tag, docker image tag
- [ ] Yes

### Breaking change label

- [ ] Yes


Thegaram:
Migrated to https://github.com/scroll-tech/scroll-contracts/pull/5.

PR #1321, opened 2024-05-05
build: update golangci-lint

lugosix:

What does this PR do?
- update golangci-lint from v1.52 to v1.57.2

Why does it do it?
- Expose problems early: Let the linter discover as many errors as possible during the compilation phase, rather than at runtime.

How does it do it?
- v1.57.2 supports a wider variety of linters

### PR title

- [x] build: Changes that affect the build system or external dependencies (example scopes: yarn, eslint, typescript)

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes

### DIFF
--- a/build/.golangci.yml
+++ b/build/.golangci.yml
@@ -104,10 +104,12 @@ linters-settings:
     # minimal occurrences count to trigger, 3 by default
     min-occurrences: 3
   depguard:
-    list-type: blacklist
-    include-go-root: false
-    packages:
-      - github.com/davecgh/go-spew/spew
+    rules:
+      main:
+        files:
+          - $all
+        deny:
+          - pkg: "github.com/davecgh/go-spew/spew"
   misspell:
     # Correct spellings using locale preferences for US or UK.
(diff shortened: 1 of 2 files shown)


build/.golangci.yml
```diff
@@ -104,10 +104,12 @@ linters-settings:
...
     min-occurrences: 3
   depguard:
-    list-type: blacklist
-    include-go-root: false
-    packages:
-      - github.com/davecgh/go-spew/spew
+    rules:
+      main:
+        files:
+          - $all
```
georgehao, line 110: Why change this?

lugosix: Starting from v1.53, there have been breaking changes introduced in the configuration. For details https://github.com/golangci/golangci-lint/discussions/3727

georgehao (approved):

build/.golangci.yml
colinlyguo: also met this when trying to upgrade it. thanks.

colinlyguo (approved):

PR #1395, opened 2024-06-24
faet(db): add bundle table

colinlyguo:

Some status enum types are shared among chunk/batch/bundle:
rollup_status: https://github.com/scroll-tech/scroll/blob/v4.4.20/common/types/db.go#L191-L211
proving_status: https://github.com/scroll-tech/scroll/blob/v4.4.20/common/types/db.go#L131-L147

### PR title

- [x] feat: A new feature

### DIFF
--- a/common/types/db.go
+++ b/common/types/db.go
@@ -188,7 +188,7 @@ func (s ChunkProofsStatus) String() string {
 	}
 }
 
-// BatchProofsStatus describes the proving status of batchs that belong to a bundle.
+// BatchProofsStatus describes the proving status of batches that belong to a bundle.
 type BatchProofsStatus int
 
 const (
--- a/common/types/db.go
+++ b/common/types/db.go
@@ -195,10 +195,10 @@ const (
 	// BatchProofsStatusUndefined represents an undefined batch proofs status
 	BatchProofsStatusUndefined BatchProofsStatus = iota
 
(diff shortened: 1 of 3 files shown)


common/types/db.go
```diff
@@ -188,6 +188,31 @@ func (s ChunkProofsStatus) String() string {
 	}
 }
 
+// BatchProofsStatus describes the proving status of batchs that belong to a bundle.
```
georgehao:
```suggestion
// BatchProofsStatus describes the proving status of batches that belong to a bundle.
```

georgehao (approved):

common/types/db.go
```diff
@@ -188,6 +188,31 @@ func (s ChunkProofsStatus) String() string {
 	}
 }
 
+// BatchProofsStatus describes the proving status of batches that belong to a bundle.
```
amoylan2, line 191: How about using "Bundle" replace "Batch" as the prefix？Since this enum is related to a field in bundle table

colinlyguo: use a same naming as https://github.com/scroll-tech/scroll/blob/v4.4.20/common/types/db.go#L167

amoylan2: I misunderstood this. No question

amoylan2 (approved):
LGTM

PR #1397, opened 2024-06-25
feat(rollup-relayer): support commitBatchWithBlobProof

colinlyguo:

- [x] respect timestamp-based hardfork in chunk-proposer and batch-proposer.

### PR title

- [x] feat: A new feature


rollup/internal/controller/watcher/batch_proposer_test.go
```diff
@@ -227,15 +216,6 @@ func testBatchProposerCodecv1Limits(t *testing.T) {
 			expectedBatchesLen:         1,
 			expectedChunksInFirstBatch: 1,
 		},
-		{
```
omerfirmak, line 230: why are we removing these tests? We should still have support for height-based forks, no?

rollup/internal/controller/watcher/chunk_proposer.go
```diff
@@ -194,10 +188,6 @@ func (p *ChunkProposer) proposeChunk() error {
 	}
 
 	maxBlocksThisChunk := p.maxBlockNumPerChunk
-	blocksUntilFork := forks.BlocksUntilFork(unchunkedBlockHeight, p.forkHeights)
-	if blocksUntilFork != 0 && blocksUntilFork < maxBlocksThisChunk {
-		maxBlocksThisChunk = blocksUntilFork
-	}
```
omerfirmak, line 200: if we remove this, a chunk may include blocks from different forks. Which is not something we want.

colinlyguo: it's replaced by the following logic below:
```
	// Ensure all blocks in the same chunk use the same codec version
	// If a different codec version is found, truncate the blocks slice at that point
	for i := 1; i < len(blocks); i++ {
		currentCodecVersion := getCodecVersion(p.chainCfg, blocks[i].Header.Number.Uint64(), blocks[i].Header.Time)
		if currentCodecVersion != codecVersion {
			blocks = blocks[:i]
			maxBlocksThisChunk = uint64(i) // update maxBlocksThisChunk to trigger chunking, because these blocks are the last blocks before the hardfork
			break
		}
	}
```

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -205,13 +184,13 @@ func (p *BatchProposer) proposeBatch() error {
 		return nil
 	}
 
+	// Ensure all chunks in the same batch use the same codec version
+	// If a different codec version is found, truncate the chunks slice at that point
 	for i, chunk := range dbChunks {
-		// if a chunk is starting at a fork boundary, only consider earlier chunks
-		if i != 0 && p.forkMap[chunk.StartBlockNumber] {
+		currentCodecVersion := getCodecVersion(p.chainCfg, chunk.StartBlockNumber, chunk.StartBlockTime)
+		if currentCodecVersion != codecVersion {
```
omerfirmak: This will only work if a fork changes the codec as well. If we ever have a fork that doesn't come with a new codec version, we might end up putting chunks of different forks in the same batch.

rollup/internal/controller/watcher/batch_proposer_test.go
colinlyguo: yeah. since in new implementations, rollup-relayer will only recognize `bernoulli`, `curie`, `darwin`, so this test is no longer compatible with the current logic, will add tests with the same intentions in an independent unit test.

rollup/internal/controller/watcher/chunk_proposer.go
omerfirmak: same concern [here](https://github.com/scroll-tech/scroll/pull/1397#discussion_r1655099604) applies to chunks as well

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: makes sense, will use hardfork name to detect if there is a hardfork between these blocks or chunks, instead of codec versions.

colinlyguo: changed to use hardfork name in this commit: [5cde900](https://github.com/scroll-tech/scroll/pull/1397/commits/5cde9004168ae9ae2fee0c59f00418ae5e6ef35d)

rollup/internal/controller/watcher/chunk_proposer.go
colinlyguo: fixed in this commit: [5cde900](https://github.com/scroll-tech/scroll/pull/1397/commits/5cde9004168ae9ae2fee0c59f00418ae5e6ef35d)

rollup/internal/controller/watcher/batch_proposer_test.go
colinlyguo: added in [887d2b1](https://github.com/scroll-tech/scroll/pull/1397/commits/887d2b1f851049a91b652206f648ca5b7302313c).

omerfirmak:
Fork handling LGTM

Thegaram (commented):
Looks good

colinlyguo:
I'll add unit tests (commitBatch first, then verify batch hash based on event) when the contract's `commitBatchWithBlobProof` is ready with another PR.

georgehao (approved):

PR #1400, opened 2024-06-26
feat(common): add hardfork util functions supporting both height and timestamp

colinlyguo:

Put these functions in a common place for the moment.

### PR title

- [x] feat: A new feature


common/forks/forks.go
```diff
@@ -87,3 +88,45 @@ func BlockRange(currentForkHeight uint64, forkHeights []uint64) (from, to uint64
...
+// It checks the chain configuration to determine which hardfork is active.
+func GetHardforkName(config *params.ChainConfig, blockHeight, blockTimestamp uint64) string {
+	if !config.IsBernoulli(new(big.Int).SetUint64(blockHeight)) {
+		return "homestead"
+	} else if !config.IsCurie(new(big.Int).SetUint64(blockHeight)) {
+		return "bernoulli"
+	} else if !config.IsDarwin(blockTimestamp) {
+		return "curie"
+	} else {
+		return "darwin"
```
omerfirmak, line 102: Can we make fork names an enum instead of string constants?

colinlyguo: yes, we can try enum type, too.
I'm using string constants because I remembered that the coordinator and prover use these string constants to communicate? cc @georgehao

georgehao: yes. need string constants

georgehao (approved):

common/forks/forks.go
omerfirmak: We could've made them string enums.

PR #1404, opened 2024-06-28
feat(bridge-history): change FinalizeBatch event logic

amoylan2:

Upgrade the logic when dealing with FinalizeBatch event. 
This is intended to support darwin as well as backward compatible.

One thing need caution: "GetLastFinalizedBatchIndex" from db can not use any index and it's not suitable to add such index since there will be almost 100% batches are all finalized as time goes. Do we need add a key-value mode cache to save the "last finalized batch index"? it can be a small kv table in postgres or using redis..

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes

### DIFF
--- a/bridge-history-api/internal/controller/fetcher/l1_fetcher.go
+++ b/bridge-history-api/internal/controller/fetcher/l1_fetcher.go
@@ -153,6 +153,7 @@ func (c *L1MessageFetcher) fetchAndSaveEvents(confirmation uint64) {
 			return
 		}
 
+		c.l1FetcherLogic.ShiftToNextLastFinalizedBatchIndex()
 		c.updateL1SyncHeight(to, lastBlockHash)
 		c.l1MessageFetcherRunningTotal.Inc()
 	}
--- a/bridge-history-api/internal/controller/fetcher/l1_fetcher.go
+++ b/bridge-history-api/internal/controller/fetcher/l1_fetcher.go
@@ -41,7 +41,7 @@ func NewL1MessageFetcher(ctx context.Context, cfg *config.FetcherConfig, db *gor
 		cfg:              cfg,
 		client:           client,
 		eventUpdateLogic: logic.NewEventUpdateLogic(db, true),
-		l1FetcherLogic:   logic.NewL1FetcherLogic(cfg, db, client),
+		l1FetcherLogic:   logic.NewL1FetcherLogic(ctx, cfg, db, client),
 	}
 
 	reg := prometheus.DefaultRegisterer
--- a/bridge-history-api/internal/logic/l1_event_parser.go
+++ b/bridge-history-api/internal/logic/l1_event_parser.go
@@ -224,7 +224,8 @@ func (e *L1EventParser) ParseL1SingleCrossChainEventLogs(ctx context.Context, lo
 }
 
 // ParseL1BatchEventLogs parses L1 watched batch events.
-func (e *L1EventParser) ParseL1BatchEventLogs(ctx context.Context, logs []types.Log, client *ethclient.Client) ([]*orm.BatchEvent, error) {
+func (e *L1EventParser) ParseL1BatchEventLogs(ctx context.Context, logs []types.Log, f *L1FetcherLogic) ([]*orm.BatchEvent, error) {
+	client := f.client
 	var l1BatchEvents []*orm.BatchEvent
 	for _, vlog := range logs {
 		switch vlog.Topics[0] {
@@ -270,12 +271,17 @@ func (e *L1EventParser) ParseL1BatchEventLogs(ctx context.Context, logs []types.
 				log.Error("Failed to unpack FinalizeBatch event", "err", err)
 				return nil, err
 			}
-			l1BatchEvents = append(l1BatchEvents, &orm.BatchEvent{
-				BatchStatus:   int(btypes.BatchStatusTypeFinalized),
-				BatchIndex:    event.BatchIndex.Uint64(),
-				BatchHash:     event.BatchHash.String(),
-				L1BlockNumber: vlog.BlockNumber,
-			})
+			var startBatchIndex uint64 = 0
+			if lastFinalizedBatchIndex, ok := f.getLastFinalizedBatchIndex(); ok {
+				startBatchIndex = lastFinalizedBatchIndex + 1
+			}
+			for index := startBatchIndex; index <= event.BatchIndex.Uint64(); index++ {
+				l1BatchEvents = append(l1BatchEvents, &orm.BatchEvent{
+					BatchStatus: int(btypes.BatchStatusTypeFinalized),
+					BatchIndex:  index,
+				})
+			}
+			f.setNextPendingLastFinalizedBatchIndex(event.BatchIndex.Uint64())
 		}
 	}
 	return l1BatchEvents, nil
--- a/bridge-history-api/internal/logic/l1_fetcher.go
+++ b/bridge-history-api/internal/logic/l1_fetcher.go
@@ -45,6 +45,9 @@ type L1FetcherLogic struct {
 	crossMessageOrm *orm.CrossMessage
 	batchEventOrm   *orm.BatchEvent
 
+	lastFinalizedBatchIndex            *uint64
+	nextPendingLastFinalizedBatchIndex uint64
+
 	l1FetcherLogicFetchedTotal *prometheus.CounterVec
 }
 
@@ -117,6 +120,12 @@ func NewL1FetcherLogic(cfg *config.FetcherConfig, db *gorm.DB, client *ethclient
 		gatewayList:     gatewayList,
 		parser:          NewL1EventParser(cfg, client),
 	}
+
+	index, err := f.batchEventOrm.GetLastFinalizedBatchIndex(context.TODO())
+	if err != nil && err != gorm.ErrRecordNotFound {
+		log.Crit("failed to get finalized bacth index, fatal db error", "err", err.Error())
+	}
+	f.setLastFinalizedBatchIndex(index)
 
 	reg := prometheus.DefaultRegisterer
 	f.l1FetcherLogicFetchedTotal = promauto.With(reg).NewCounterVec(prometheus.CounterOpts{
@@ -266,7 +275,7 @@ func (f *L1FetcherLogic) L1Fetcher(ctx context.Context, from, to uint64, lastBlo
 		return false, 0, common.Hash{}, nil, err
 	}
 
-	l1BatchEvents, err := f.parser.ParseL1BatchEventLogs(ctx, eventLogs, f.client)
+	l1BatchEvents, err := f.parser.ParseL1BatchEventLogs(ctx, eventLogs, f)
 	if err != nil {
 		log.Error("failed to parse L1 batch event logs", "from", from, "to", to, "err", err)
 		return false, 0, common.Hash{}, nil, err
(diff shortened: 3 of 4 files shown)


bridge-history-api/internal/orm/batch_event.go
```diff
@@ -89,8 +89,9 @@ func (c *BatchEvent) InsertOrUpdateBatchEvents(ctx context.Context, l1BatchEvent
 				return fmt.Errorf("failed to insert or ignore batch event, error: %w", err)
 			}
 		case btypes.BatchStatusTypeFinalized:
+			// After darwin, FinalizeBatch event signals a range of batches are finalized,
+			// thus losing the batch hash info. Meanwhile, only batch_index is enough to find the target batch.
```
colinlyguo:
```suggestion
			// thus losing the batch hash info. Meanwhile, only batch_index is enough to find the target batch when finalizing batches.
```

colinlyguo:
This feature can be directly merged into the develop branch since it's backward compatible, we can deploy it first.

For discussion about: _One thing need caution: "GetLastFinalizedBatchIndex" from db can not use any index and it's not suitable to add such index since there will be almost 100% batches are all finalized as time goes. Do we need add a key-value mode cache to save the "last finalized batch index"? it can be a small kv table in postgres or using redis.._

this can be mitigated by the "composite index" as follows:
```
CREATE INDEX IF NOT EXISTS idx_be_batch_index_batch_status ON batch_event_v2 (batch_index, batch_status);
```

but I think using `batch_index` is sufficient since in your newly added query, it's sorted by `batch_index desc`, there won't be too many iterations before finding the first finalized batch.

bridge-history-api/internal/logic/l1_event_parser.go
```diff
@@ -270,12 +271,17 @@ func (e *L1EventParser) ParseL1BatchEventLogs(ctx context.Context, logs []types.
...
-				BatchStatus:   int(btypes.BatchStatusTypeFinalized),
-				BatchIndex:    event.BatchIndex.Uint64(),
-				BatchHash:     event.BatchHash.String(),
-				L1BlockNumber: vlog.BlockNumber,
-			})
+			var startBatchIndex uint64 = 0
+			if lastFinalizedBatchIndex, ok := f.getLastFinalizedBatchIndex(); ok {
+				startBatchIndex = lastFinalizedBatchIndex + 1
+			}
+			for index := startBatchIndex; index <= event.BatchIndex.Uint64(); index++ {
```
colinlyguo, line 278: What about changing the pattern of SQL update to <= batch index, and only keeping the last event (with max batch index)? This can reduce the number of SQL update ops.

bridge-history-api/internal/logic/l1_fetcher.go
```diff
@@ -118,6 +121,12 @@ func NewL1FetcherLogic(cfg *config.FetcherConfig, db *gorm.DB, client *ethclient
 		parser:          NewL1EventParser(cfg, client),
 	}
 
+	index, err := f.batchEventOrm.GetLastFinalizedBatchIndex(context.TODO())
```
colinlyguo: passing a context from the parent call stack here when initializing the l1_fetcher instance?

bridge-history-api/internal/orm/batch_event.go
```diff
@@ -124,3 +125,15 @@ func (c *BatchEvent) UpdateBatchEventStatus(ctx context.Context, batchIndex uint
...
+
+// GetLastFinalizedBatchIndex returns the last finalized batch index in db.
+func (c *BatchEvent) GetLastFinalizedBatchIndex(ctx context.Context) (uint64, error) {
+	var batche BatchEvent
+	db := c.db.WithContext(ctx)
+	db = db.Model(&BatchEvent{})
+	db = db.Select("batch_index")
+	db = db.Where("batch_status = ?", btypes.BatchStatusTypeFinalized)
+	db = db.Order("batch_index desc")
+	err := db.First(&batche).Error
```
colinlyguo, line 137: suggesting handling `not found` explicitly, and returning 0 here when `not found` for readability.

bridge-history-api/internal/controller/fetcher/l1_fetcher.go
```diff
@@ -153,6 +153,7 @@ func (c *L1MessageFetcher) fetchAndSaveEvents(confirmation uint64) {
 			return
 		}
 
+		c.l1FetcherLogic.ShiftToNextLastFinalizedBatchIndex()
```
colinlyguo, line 156: I think updating the status here is enough. reorg won't affect this info. pls revisit this logic a bit. @amoylan2

amoylan2:
I will open another PR with target branch "develop".
If the other is merged this can be closed anytime

bridge-history-api/internal/logic/l1_fetcher.go
```diff
@@ -45,6 +45,9 @@ type L1FetcherLogic struct {
 	crossMessageOrm *orm.CrossMessage
 	batchEventOrm   *orm.BatchEvent
 
+	lastFinalizedBatchIndex            *uint64
+	nextPendingLastFinalizedBatchIndex uint64
```
colinlyguo, line 49: is this design like pre-commit, then commit? could these two variables be merged into one?

colinlyguo (approved):

bridge-history-api/internal/logic/l1_fetcher.go
amoylan2: `lastFinalizedBatchIndex` should be updated only after the events are inserted or updated to the db in which situation the context info of new value is lost.
So I add another value as a temporary store media

bridge-history-api/internal/controller/fetcher/l1_fetcher.go
amoylan2: The logic in line 156 is intended to make the update after the `L1InsertOrUpdate` method, where `batch_event_v2` table is updated.
I think it has nothing to do with the reorg thing

amoylan2:
> This feature can be directly merged into the develop branch since it's backward compatible, we can deploy it first.
> 
> For discussion about: _One thing need caution: "GetLastFinalizedBatchIndex" from db can not use any index and it's not suitable to add such index since there will be almost 100% batches are all finalized as time goes. Do we need add a key-value mode cache to save the "last finalized batch index"? it can be a small kv table in postgres or using redis.._
> 
> this can be mitigated by the "composite index" as follows:
> 
> ```
> CREATE INDEX IF NOT EXISTS idx_be_batch_index_batch_status ON batch_event_v2 (batch_index, batch_status);
> ```
> 
> but I think using `batch_index` is sufficient since in your newly added query, it's sorted by `batch_index desc`, there won't be too many iterations before finding the first finalized batch.

shall we discuss in new pr ? or discuss here first

bridge-history-api/internal/logic/l1_event_parser.go
amoylan2: I'm a little confused of your comment here. 
The code snippet has nothing to do with any sql action. The `getLastFinalizedBatchIndex` method is already a local cache.
The only sql action is needed when initializing the `L1FetcherLogic` struct

bridge-history-api/internal/logic/l1_fetcher.go
amoylan2: ok, will do that

colinlyguo:
> > This feature can be directly merged into the develop branch since it's backward compatible, we can deploy it first.
> > For discussion about: _One thing need caution: "GetLastFinalizedBatchIndex" from db can not use any index and it's not suitable to add such index since there will be almost 100% batches are all finalized as time goes. Do we need add a key-value mode cache to save the "last finalized batch index"? it can be a small kv table in postgres or using redis.._
> > this can be mitigated by the "composite index" as follows:
> > ```
> > CREATE INDEX IF NOT EXISTS idx_be_batch_index_batch_status ON batch_event_v2 (batch_index, batch_status);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > but I think using `batch_index` is sufficient since in your newly added query, it's sorted by `batch_index desc`, there won't be too many iterations before finding the first finalized batch.
> 
> shall we discuss in new pr ? or discuss here first

both LGTM.

bridge-history-api/internal/orm/batch_event.go
amoylan2: I have dealt with the `not found` error at the caller of this method.
Except the `not found` error, there may other error types which will are not easy to handle in `inner` method.
In this case, the caller will panic (log.Crit) when encountering other errors since there seems no better way to handle it..
And it would be a little weird if the `inner` function only handles `not found` and return other type errors.
In conclusion I think this `inner` function should just return error and let the caller to deal with it.

bridge-history-api/internal/logic/l1_event_parser.go
colinlyguo: I mean we can update the SQL like:
```
		case btypes.BatchStatusTypeFinalized:
			// After darwin, FinalizeBatch event signals a range of batches are finalized,
			// thus losing the batch hash info. Meanwhile, only batch_index is enough to find the target batch when finalizing batches.
			db = db.Where("batch_index <= ?", l1BatchEvent.BatchIndex)
			db = db.Where("batch_status != ?", btypes.BatchStatusTypeFinalized)
			updateFields["batch_status"] = btypes.BatchStatusTypeFinalized
			if err := db.Updates(updateFields).Error; err != nil {
				return fmt.Errorf("failed to update batch event, error: %w", err)
			}
```

bridge-history-api/internal/logic/l1_fetcher.go
```diff
@@ -118,6 +121,16 @@ func NewL1FetcherLogic(cfg *config.FetcherConfig, db *gorm.DB, client *ethclient
 		parser:          NewL1EventParser(cfg, client),
 	}
 
+	index, err := f.batchEventOrm.GetLastFinalizedBatchIndex(context.TODO())
+	if err == gorm.ErrRecordNotFound {
+		// do nothing, leaving lastFinalizedBatchIndex to be nil
+		log.Warn("all bacthes are non-finalized, this should happen only once")
+	} else if err != nil {
+		log.Crit("failed to get last finalized bacth index, fatal db error", "err", err.Error())
```
colinlyguo, line 129: we can throw the other types of errors and exit this round instead of panicking here. some could succeed after retrying.

bridge-history-api/internal/logic/l1_fetcher.go
colinlyguo: oh I see. looks like needing at least two variables to prevent early return before flushing new height to `lastFinalizedBatchIndex`.

bridge-history-api/internal/logic/l1_event_parser.go
amoylan2: got it, let me check

bridge-history-api/internal/logic/l1_fetcher.go
amoylan2: let me check it

bridge-history-api/internal/logic/l1_event_parser.go
amoylan2: literally one BatchEvent represents a single record in the database which relates to a single batch.
I think it's better to keep the BatchEvent array compared to change the BatchEvent meaning.

When it comes to the db operations, it can be collected the boundaries of batch indexes before doing the update. Besides, doing a range update with `<=` in a for loop is not efficient anyway.

colinlyguo: Doing a range update with <= in a for loop is not efficient anyway.
because after Darwin, there may be several batches per update, if we use `batch_index <= a && batch_status != finalized`, this can reduce the number of RPCs.
yeah. it's low frequency compared with cross-chain messages updates, but previously this was a very useful optimization (we merged the one-by-one read queries into one, reducing rpc latencies).

colinlyguo: _Doing a range update with <= in a for loop is not efficient anyway._
because after Darwin, there may be several batches per update, if we use `batch_index <= a && batch_status != finalized`, this can reduce the number of RPCs.
yeah. it's low frequency compared with cross-chain messages updates, but previously this was a very useful optimization (we merged the one-by-one read queries into one, reducing rpc latencies).

the new approach:
1. only store the FinalizeBatch event (one time), instead of splitting it into multiple events.
2. update db once.

amoylan2: I have a new implementation, pls have a look

bridge-history-api/internal/logic/l1_fetcher.go
amoylan2: job is done

bridge-history-api/internal/logic/l1_fetcher.go
```diff
@@ -45,11 +45,14 @@ type L1FetcherLogic struct {
 	crossMessageOrm *orm.CrossMessage
 	batchEventOrm   *orm.BatchEvent
 
+	lastFinalizedBatchIndex            *uint64
+	nextPendingLastFinalizedBatchIndex uint64
```
colinlyguo, line 49: we can remove them if using `batch_index <= x && batch_status != finalized` to update db.

bridge-history-api/internal/logic/l1_fetcher.go
amoylan2: it's in the process initializing stage, so let's panic it

bridge-history-api/internal/logic/l1_fetcher.go
colinlyguo: finalization can be treated as deterministic.

PR #1414, opened 2024-07-08
fix(rollup-relayer): avoid compressing "small tip" block

colinlyguo:

Applying related fix: https://github.com/scroll-tech/da-codec/pull/17

- [x] Updating the dependency in `go.mod` after merging the above pr.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label


Thegaram (commented):
Need to consider the error from `ConstructBlobPayload` in `batch-proposer`.

rollup/internal/controller/watcher/chunk_proposer_test.go
```diff
@@ -568,51 +568,53 @@ func testChunkProposerCodecv2Limits(t *testing.T) {
...
+			chainConfig = &params.ChainConfig{}
+		} else if codecVersion == encoding.CodecV1 {
 			chainConfig = &params.ChainConfig{BernoulliBlock: big.NewInt(0)}
+		} else {
+			chainConfig = &params.ChainConfig{BernoulliBlock: big.NewInt(0), CurieBlock: big.NewInt(0)}
 		}
 
 		cp := NewChunkProposer(context.Background(), &config.ChunkProposerConfig{
-			MaxBlockNumPerChunk:             math.MaxUint64,
+			MaxBlockNumPerChunk:             255,
```
colinlyguo, line 592: da-codec has added a max 255 blocks per chunk check, thus fixing this unit test.

rollup/internal/utils/utils.go
```diff
@@ -91,7 +93,12 @@ func CalculateChunkMetrics(chunk *encoding.Chunk, codecVersion encoding.CodecVer
 		metrics.L1CommitUncompressedBatchBytesSize, metrics.L1CommitBlobSize, err = codecv2.EstimateChunkL1CommitBatchSizeAndBlobSize(chunk)
 		metrics.EstimateBlobSizeTime = time.Since(start)
 		if err != nil {
-			return nil, fmt.Errorf("failed to estimate codecv2 chunk L1 commit blob size: %w", err)
+			if errors.Is(err, &encoding.CompressedDataCompatibilityError{}) {
+				log.Error("Compressed data compatibility check failed", "error", err)
```
Thegaram: Why log an error here?

First of all this is already logger in `da-codec`.

Second, we don't know here if this is an error. We'll inspect and log accordingly in higher layers (chunk-proposer, etc.)

Thegaram (commented):
LGTM, let's update deps once the other PR is merged.

rollup/internal/utils/utils.go
colinlyguo: remove the `log.Error` and add `log.Warn` logs both in `chunk-proposer` and `batch-proposer` to print more info to track the chunk & batch proposed due to violation.

colinlyguo:
> LGTM, let's update deps once the other PR is merged.

updated.

colinlyguo:
some issues in unit tests, debugging.

georgehao (approved):

Thegaram (approved):

PR #1415, opened 2024-07-08
feat(bridge-history): generate withdraw proofs by bundle

colinlyguo:

This PR adds generating withdraw proofs by bundle logic, and it's backward compatible.

Notes:
It keeps commit/revert batches to maintain block range of the batch, because in `FinalizeBatch` event, there isn't a block range.

Only updating batch's status as `finalized` when listening to the `FinalizeBatch` event (when "finalize by bundle", only each bundle's last batch will be updated as `finalized`), so that bridge-history-backend can track the block range to update regardless "finalize by batch" or "finalize by bundle".

This PR is compatible with both "finalize by batch" and "finalize by bundle" modes:
- In "finalize by batch" mode, each batch emits a FinalizedBatch event.
- In "finalize by bundle" mode, all batches in the bundle emit only one `FinalizedBatch` event, using the last batch's index and hash.

For withdraw proof generation:
- finalize by batch: Generates proofs for each batch.
- finalize by bundle: Generates proofs for the entire bundle at once.

For batch index updating:
- finalize by batch: Updates the batch index for withdrawal messages in each processed batch.
- finalize by bundle: Updates the batch index for all withdrawal messages in the bundle, using the index of the last batch in the bundle.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [x] Yes

### Breaking change label

### DIFF
--- a/bridge-history-api/conf/config.json
+++ b/bridge-history-api/conf/config.json
@@ -19,7 +19,7 @@
 		"ScrollChainAddr": "0xa13BAF47339d63B743e7Da8741db5456DAc1E556",
 		"GatewayRouterAddr": "0xF8B1378579659D8F7EE5f3C929c2f3E332E41Fd6",
 		"MessageQueueAddr": "0x0d7E906BD9cAFa154b048cFa766Cc1E54E39AF9B",
-		"BatchBridgeGatewayAddr": "0x0000000000000000000000000000000000000000"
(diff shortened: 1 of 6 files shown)


zimpha (dismissed):

georgehao (dismissed):

georgehao:
change the target branch

zimpha (approved):

georgehao (approved):

PR #1425, opened 2024-07-10
Delete dead badge

FredPhilipy:
Sorry for the bothering, but it looks like your homepage(README) badge is down due to [PR](https://github.com/scroll-tech/scroll/commit/ab9c5414093082141c55ff562899234462a1e7e3#diff-d0024dcfce8263455f76f35323f55051631264ba689777e927397480a6914ff4), if this action is not used anymore, I think we should delete it?

image


Thegaram (approved):

Thegaram:
Thanks!

PR #1431, opened 2024-07-11
feat(chunk&batch proposers): check compressed data validity

colinlyguo:

Since RAW/RLE blocks are not highly related to chunk size and batch size, this PR moves the compressed data validity check after chunks and batches are ready.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [x] Yes

### Breaking change label


0xmountaintop (dismissed):

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -155,6 +155,28 @@ func (p *BatchProposer) TryProposeBatch() {
 }
 
 func (p *BatchProposer) updateDBBatchInfo(batch *encoding.Batch, codecVersion encoding.CodecVersion, metrics utils.BatchMetrics) error {
+	for {
```
noel2004, line 160: So we would come into this loop only when we have enough chunks in the batch, is it right?

Currently, which factors would constraint the maxium number of chunks?

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -155,6 +155,28 @@ func (p *BatchProposer) TryProposeBatch() {
...
+		compatible, err := utils.CheckBatchCompressedDataCompatibility(batch, codecVersion)
+		if err != nil {
+			log.Error("Failed to check batch compressed data compatibility", "err", err)
```
georgehao:
```suggestion
			log.Error("Failed to check batch compressed data compatibility", "codecVersion", codecVersion,  "err", err)
```

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -155,6 +155,28 @@ func (p *BatchProposer) TryProposeBatch() {
...
+			break
+		}
+
+		if len(batch.Chunks) == 1 {
+			log.Error("Cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
+				"end block number", batch.Chunks[0].Blocks[len(batch.Chunks[0].Blocks)-1].Header.Number.Uint64())
+			return errors.New("cannot truncate batch with only 1 chunk for compatibility")
+		}
+
+		log.Warn("Batch not compatible with compressed data, removing last chunk")
```
georgehao: `warn` level may be not good. I think this a normal logic

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: yes.

Blob size, uncompressed batch size, first block's timestamp, calldata size and commit tx's gas limit will all be treated as constraints of proposing a batch.

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: makes sense. changed in [0944e49](https://github.com/scroll-tech/scroll/pull/1431/commits/0944e498da4a1a2f6b785bf1849202819a0335f0).

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: added this info in [0944e49](https://github.com/scroll-tech/scroll/pull/1431/commits/0944e498da4a1a2f6b785bf1849202819a0335f0).

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -155,6 +155,28 @@ func (p *BatchProposer) TryProposeBatch() {
...
+			break
+		}
+
+		if len(batch.Chunks) == 1 {
+			log.Error("Cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
+				"end block number", batch.Chunks[0].Blocks[len(batch.Chunks[0].Blocks)-1].Header.Number.Uint64())
+			return errors.New("cannot truncate batch with only 1 chunk for compatibility")
+		}
+
+		log.Info("Batch not compatible with compressed data, removing last chunk")
```
Thegaram: I recommend adding more info here, maybe the batch index and the number of chunks?

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -155,6 +155,28 @@ func (p *BatchProposer) TryProposeBatch() {
...
+		p.compressedDataCompatibilityBreachTotal.Inc()
+		batch.Chunks = batch.Chunks[:len(batch.Chunks)-1]
+	}
+
 	err := p.db.Transaction(func(dbTX *gorm.DB) error {
 		dbBatch, dbErr := p.batchOrm.InsertBatch(p.ctx, batch, codecVersion, metrics, dbTX)
```
Thegaram: `metrics` will be incorrect

Thegaram (changes requested):

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: I considered this when implementing, and after revisiting, submitting only once makes more sense. will change it.

colinlyguo: bug fixed in [85f995b](https://github.com/scroll-tech/scroll/pull/1431/commits/85f995ba36832342801724b92fe0aaa06b180646).

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: added more info in [85f995b](https://github.com/scroll-tech/scroll/pull/1431/commits/85f995ba36832342801724b92fe0aaa06b180646).

rollup/internal/controller/watcher/batch_proposer.go
Thegaram: No that's not my point. This `metrics` will be incorrect:

```
p.batchOrm.InsertBatch(p.ctx, batch, codecVersion, metrics, dbTX)
```

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -154,9 +154,47 @@ func (p *BatchProposer) TryProposeBatch() {
...
+	if compatibilityBreachOccurred {
+		p.compressedDataCompatibilityBreachTotal.Inc()
+	}
+
+	// calculate metrics before updating the batch info in db
+	metrics, err := utils.CalculateBatchMetrics(batch, codecVersion)
+	if err != nil {
+		return fmt.Errorf("failed to calculate batch metrics: %w", err)
+	}
+	p.recordTimerBatchMetrics(metrics)
```
colinlyguo: also removes `recordTimerBatchMetrics` in other places, these metrics are mainly used to reflect zstd compression time, we can record it when the batch is ready, omitting the trying-out phase (batch is usually smaller).

so does `recordTimerChunkMetrics`.

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: nice catch. fixed in [9b7f048](https://github.com/scroll-tech/scroll/pull/1431/commits/9b7f048b93100f3712e6315a5828584729af4536).

the commit also simplified the related logic a bit.

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -154,9 +154,47 @@ func (p *BatchProposer) TryProposeBatch() {
...
-func (p *BatchProposer) updateDBBatchInfo(batch *encoding.Batch, codecVersion encoding.CodecVersion, metrics utils.BatchMetrics) error {
-	err := p.db.Transaction(func(dbTX *gorm.DB) error {
-		dbBatch, dbErr := p.batchOrm.InsertBatch(p.ctx, batch, codecVersion, metrics, dbTX)
+func (p *BatchProposer) updateDBBatchInfo(batch *encoding.Batch, codecVersion encoding.CodecVersion) error {
+	compatibilityBreachOccurred := false
+
+	for {
+		compatible, err := utils.CheckBatchCompressedDataCompatibility(batch, codecVersion)
+		if err != nil {
+			log.Error("Failed to check batch compressed data compatibility", "codecVersion", codecVersion, "err", err)
```
Thegaram: log which batch

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -154,9 +154,47 @@ func (p *BatchProposer) TryProposeBatch() {
...
+			return err
+		}
+
+		if compatible {
+			break
+		}
+
+		if len(batch.Chunks) == 1 {
+			log.Error("Cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
```
Thegaram, line 172: Why did you decide to log the block numbers, as opposed to the chunk index?

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -154,9 +154,47 @@ func (p *BatchProposer) TryProposeBatch() {
...
+
+	// calculate metrics before updating the batch info in db
+	metrics, err := utils.CalculateBatchMetrics(batch, codecVersion)
+	if err != nil {
+		return fmt.Errorf("failed to calculate batch metrics: %w", err)
+	}
+	p.recordTimerBatchMetrics(metrics)
+	p.recordAllBatchMetrics(metrics)
+
+	p.proposeBatchUpdateInfoTotal.Inc()
```
Thegaram, line 197: This was missing before?

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -239,25 +277,10 @@ func (p *BatchProposer) proposeBatch() error {
...
-			batch.Chunks = batch.Chunks[:len(batch.Chunks)-1]
-			p.compressedDataCompatibilityBreachTotal.Inc()
-			return p.updateDBBatchInfo(&batch, codecVersion, *metrics)
-		}
-
 		if calcErr != nil {
 			return fmt.Errorf("failed to calculate batch metrics: %w", calcErr)
 		}
 
-		p.recordTimerBatchMetrics(metrics)
```
Thegaram: We don't want to collect timer metrics for all invocations of `CalculateBatchMetrics`?

rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -279,14 +302,7 @@ func (p *BatchProposer) proposeBatch() error {
...
 			batch.Chunks = batch.Chunks[:len(batch.Chunks)-1]
-
-			metrics, err := utils.CalculateBatchMetrics(&batch, codecVersion)
-			if err != nil {
-				return fmt.Errorf("failed to calculate batch metrics: %w", err)
-			}
-
-			p.recordAllBatchMetrics(metrics)
-			return p.updateDBBatchInfo(&batch, codecVersion, *metrics)
+			return p.updateDBBatchInfo(&batch, codecVersion)
```
Thegaram: This is acceptable, but in the happy path (when the compatibility check passes), we could just use `metrics` from here. In your current version, we need to run `CalculateBatchMetrics` 2x for each batch.

Thegaram (dismissed):

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: emmm. chunk index should be fetched from db. it's not included in `encoding.Chunk`.
by using start block number, we can also find the chunk (if it is successfully proposed and inserted into db).

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: Yes.

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: that's also what I'm thinking of. which results do you think are more valuable for analysis?

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: this info is quite helpful.

rollup/internal/controller/watcher/batch_proposer.go
colinlyguo: reverted some changes to avoid x2 metric calculation. and this missing metric is kept.

Thegaram (approved):

georgehao (approved):

PR #1441, opened 2024-07-15
feat: do not update fee if not committing batches

NazariiDenha:

[task](https://www.notion.so/scrollzkp/gas-oracle-do-not-update-L2-fees-if-we-re-not-committing-batches-8254e5a659ec4f3ba17e0442fdf71276?pvs=4)

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] Yes


rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -230,6 +234,20 @@ func (r *Layer1Relayer) StopSenders() {
...
+		"rollup_status": types.RollupCommitting,
+	}
+	orderByList := []string{"updated_at DESC"}
+	limit := 1
+	batches, err := r.batchOrm.GetBatches(r.ctx, fields, orderByList, limit)
+	if err != nil {
+		log.Error("failed to fetch latest committing batches", "err", err)
+	}
+	// do not update GasOracle if we haven't committed batches for a while
+	if len(batches) == 0 || utils.NowUTC().Sub(batches[0].UpdatedAt) > time.Duration(r.cfg.GasOracleConfig.GasPriceOracleUpdateWindow)*time.Minute {
```
Thegaram: We should remove `len(batches) == 0`, we'd also match that if we successfully committed all pending batches.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -230,6 +234,20 @@ func (r *Layer1Relayer) StopSenders() {
 }
 
 func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64, isCurie bool) bool {
+	fields := map[string]interface{}{
+		"rollup_status": types.RollupCommitting,
```
Thegaram: Not clear what's the best solution here, but I think checking the update time for the latest `types.RollupCommitted` entry might be better. I.e. the last successful rollup tx that we sent to L1.

Also, we should consider updating the gas oracle in this case to some default value, otherwise it might get stuck at an expensive setting.

rollup/conf/config.json
```diff
@@ -19,7 +19,8 @@
         "min_gas_price": 0,
         "gas_price_diff": 50000,
         "l1_base_fee_weight": 0.132,
-        "l1_blob_base_fee_weight":  0.145
+        "l1_blob_base_fee_weight":  0.145,
+        "gas_price_oracle_update_window": 5
```
Thegaram: We should pick a more intuitive name, e.g.

```suggestion
        "commit_batch_timeout_window_minutes": 5
```

rollup/internal/controller/relayer/l1_relayer.go
NazariiDenha: Added default `baseFee` and `blobBaseFee` to configurations, but any suggestions what default value should be chosen?

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -259,3 +272,16 @@ func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64
 
 	return false
 }
+
+func (r *Layer1Relayer) commitBatchReachTimeout() bool {
+	fields := map[string]interface{}{
+		"rollup_status": types.RollupCommitted,
+	}
+	orderByList := []string{"updated_at DESC"}
```
colinlyguo: better use `committed_at` field which is more accurate, updated_at could be affected by later changes, e.g., updating proof status.
an example using `committed_at`:
```
utils.NowUTC().Sub(*batch.CommittedAt) > time.Duration(r.cfg.FinalizeBatchWithoutProofTimeoutSec)*time.Second
```

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -259,3 +272,16 @@ func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64
 
 	return false
 }
+
+func (r *Layer1Relayer) commitBatchReachTimeout() bool {
+	fields := map[string]interface{}{
+		"rollup_status": types.RollupCommitted,
```
colinlyguo: could select from "committed", "finalizing", "finalized", because if the spike lasts for a very long time, it would be possible that all committed batches have been finalized.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -259,3 +272,16 @@ func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64
...
+	}
+	orderByList := []string{"updated_at DESC"}
+	limit := 1
+	batches, err := r.batchOrm.GetBatches(r.ctx, fields, orderByList, limit)
```
colinlyguo, line 287: the SQL may be slow (not sure if there is a default index added by gorm in `updated_at` field cc @georgehao ), suggesting queries like `order by index desc limit 1`, utilizing the primary key, the batch with latest index in status set (committed, finalizing, finalized) surely has the largest committed time, because batches are to be committed one by one.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -150,6 +154,15 @@ func (r *Layer1Relayer) ProcessGasPriceOracle() {
 		}
 
 		if r.shouldUpdateGasOracle(baseFee, blobBaseFee, isCurie) {
+			// if we are not committing batches due to high fees then we shouldn't update fees to prevent users from paying high l1_data_fee
+			// also, set fees to some default value, because we have already updated fees to some high values, probably
+			if r.commitBatchReachTimeout() {
```
colinlyguo: It's nice to have a unit test covering this path (update gas oracle price with the default `basefee` and `blobbasefee`). could set the timeout minute to 0, preventing long waiting time.

colinlyguo (dismissed):

rollup/conf/config.json
```diff
@@ -19,7 +19,10 @@
         "min_gas_price": 0,
         "gas_price_diff": 50000,
         "l1_base_fee_weight": 0.132,
-        "l1_blob_base_fee_weight":  0.145
+        "l1_blob_base_fee_weight":  0.145,
+        "commit_batch_timeout_window_minutes": 5,
```
zimpha: can we change this to another name? This name seems to indicate the timeout to commit a batch (i.e. we should commit batch every `commit_batch_timeout_window_minutes`).

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -260,3 +273,17 @@ func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64
 
 	return false
 }
+
+func (r *Layer1Relayer) commitBatchReachTimeout() bool {
+	fields := map[string]interface{}{
+		"rollup_status IN ?": []types.RollupStatus{types.RollupCommitted, types.RollupFinalizing, types.RollupFinalized},
+	}
+	orderByList := []string{"index DESC"}
```
georgehao, line 285: it's better put the sql to batch orm file.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -260,3 +273,17 @@ func (r *Layer1Relayer) shouldUpdateGasOracle(baseFee uint64, blobBaseFee uint64
...
+	limit := 1
+	batches, err := r.batchOrm.GetBatches(r.ctx, fields, orderByList, limit)
+	if err != nil {
+		log.Error("failed to fetch latest committed, finalizing or finalized batch", "err", err)
```
georgehao: why here don't return error

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -150,6 +154,15 @@ func (r *Layer1Relayer) ProcessGasPriceOracle() {
 		}
 
 		if r.shouldUpdateGasOracle(baseFee, blobBaseFee, isCurie) {
+			// if we are not committing batches due to high fees then we shouldn't update fees to prevent users from paying high l1_data_fee
```
georgehao:
```suggestion
			// It indicates the commit/finalize batch has been stuck for a long time, it's likely that the L1 gas fee spiked.
			// If we are not committing batches due to high fees then we shouldn't update fees to prevent users from paying high l1_data_fee
```

rollup/internal/controller/relayer/l1_relayer.go
georgehao: How about add `failure_type` to `PendingTransaction`? when the sender fails, record the failure type. 

So, `commitBatchReachTimeout` reached, and also check the latest `PendingTransaction` failure type, if the failure type is `HighFee`, then reset the `baseFee, blobBaseFee`

rollup/internal/controller/relayer/l1_relayer.go
NazariiDenha: it's just preparation of args for `batchOrm.GetBatches`.  It is used the same way in other places

rollup/internal/controller/relayer/l1_relayer.go
NazariiDenha: I added a unit test
I am not sure if it is possible to determine that transaction failed due to high fees. If I understand correctly:
if tx is not confirmed during some time it's fee just increased and tx resubmitted.

colinlyguo (dismissed):

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -150,6 +154,19 @@ func (r *Layer1Relayer) ProcessGasPriceOracle() {
 		}
 
 		if r.shouldUpdateGasOracle(baseFee, blobBaseFee, isCurie) {
+			// It indicates the commit/finalize batch has been stuck for a long time, it's likely that the L1 gas fee spiked.
```
colinlyguo:
```suggestion
			// It indicates the committing batch has been stuck for a long time, it's likely that the L1 gas fee spiked.
```

colinlyguo: One risk: if committing is stuck for other reasons (e.g. bugs in the rollup-relayer), then the user will be charged by "default base fee and blob base fee" during halting (for newly generated batches).

Maybe setting the default value as Ethereum's average in the deployment's config.json could prevent fund loss in the above case.

colinlyguo: or we can query L1 gas price to see if it has a very high base fee or blob base fee.

rollup/tests/gas_oracle_test.go
```diff
@@ -101,13 +130,141 @@ func testImportL1GasPriceAfterCurie(t *testing.T) {
...
+	// relay gas price
+	l1Relayer.ProcessGasPriceOracle()
+	blocks, err = l1BlockOrm.GetL1Blocks(context.Background(), map[string]interface{}{"number": latestBlockHeight})
+	assert.NoError(t, err)
+	assert.Equal(t, len(blocks), 1)
+	assert.NotEmpty(t, blocks[0].OracleTxHash)
+	assert.Equal(t, types.GasOracleStatus(blocks[0].GasOracleStatus), types.GasOracleImporting)
+}
+
+func testImportDefaultL1GasPrice(t *testing.T) {
```
colinlyguo:
```suggestion
func testImportDefaultL1GasPriceDueToL1GasPriceSpike(t *testing.T) {
```

rollup/tests/gas_oracle_test.go
```diff
@@ -101,13 +130,141 @@ func testImportL1GasPriceAfterCurie(t *testing.T) {
...
+	time.Sleep(1 * time.Second)
+
 	// relay gas price
+	// gas price will be relayed to some default value because we didn't commit batches for a l1CfgCopy.RelayerConfig.GasOracleConfig.CheckCommittedBatchesWindowMinutes = 0 minutes
 	l1Relayer.ProcessGasPriceOracle()
 	blocks, err = l1BlockOrm.GetL1Blocks(context.Background(), map[string]interface{}{"number": latestBlockHeight})
 	assert.NoError(t, err)
 	assert.Equal(t, len(blocks), 1)
 	assert.NotEmpty(t, blocks[0].OracleTxHash)
 	assert.Equal(t, types.GasOracleStatus(blocks[0].GasOracleStatus), types.GasOracleImporting)
```
colinlyguo, line 244: also check the base_fee and blob_base_fee fields fetched by `l1BlockOrm`, to see if they are equal to the default values.

zimpha (dismissed):

rollup/tests/gas_oracle_test.go
colinlyguo: update: these default values are not stored in the database.

colinlyguo (dismissed):

colinlyguo (approved):

zimpha (approved):

PR #1484, opened 2024-08-18
feat: darwinv2

colinlyguo:

`IsDarwinV2` controls whether or not to use `codecv4`.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [x] Yes

### Breaking change label


rollup/internal/controller/watcher/batch_proposer.go
```diff
@@ -164,13 +164,15 @@ func (p *BatchProposer) updateDBBatchInfo(batch *encoding.Batch, codecVersion en
 			break
 		}
 
+		compatibilityBreachOccurred = true
+
 		if len(batch.Chunks) == 1 {
-			log.Error("Cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
+			log.Warn("Disable encode: cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
```
zimpha:
```suggestion
			log.Warn("Disable compression: cannot truncate batch with only 1 chunk for compatibility", "start block number", batch.Chunks[0].Blocks[0].Header.Number.Uint64(),
```

rollup/conf/config.json
```diff
@@ -66,6 +66,7 @@
       "l1_commit_gas_limit_multiplier": 1.2
     },
     "chunk_proposer_config": {
+      "enable_conditional_compress": false,
```
georgehao: why need an option to control it?  Isn't upgrading to codev4 by default?

colinlyguo: so that we can deploy rollup-relayer for backward compatibility first (if later PRs are to be deployed first), then reenable this config after running stable.

georgehao: I don't like this way, maybe we need a fork name and decide the hard fork time.

colinlyguo: makes sense. can discuss a formal way to schedule an upgrade (besides `genesis.json`) and change this placeholder.
also cc @zimpha

colinlyguo: removed. and use `DarwinV2` in this commit: [8fead44](https://github.com/scroll-tech/scroll/pull/1484/commits/8fead4470fe44ca447b35967a3e9f14de9b916b2).

georgehao (dismissed):

georgehao (approved):

PR #1498, opened 2024-08-26
fix(gas-oracle): excluded genesis batch when check commit batch timeout

yiweichi:

In gas oracle, when it comes to function `commitBatchReachTimeout`, there will be a `nil pointer dereference` error if we get the genesis batch. Because genesis batch doesn't have `CommittedAt` field.
So we skip the `commitBatchReachTimeout` check if it a genesis batch.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

### DIFF
--- a/common/version/version.go
+++ b/common/version/version.go
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.48"
+var tag = "v4.4.49"
 
 var commit = func() string {
 	if info, ok := debug.ReadBuildInfo(); ok {
--- a/rollup/internal/controller/relayer/l1_relayer.go
+++ b/rollup/internal/controller/relayer/l1_relayer.go
@@ -290,5 +290,6 @@ func (r *Layer1Relayer) commitBatchReachTimeout() (bool, error) {
 		return false, err
 	}
 	// len(batches) == 0 probably shouldn't ever happen, but need to check this
-	return len(batches) == 0 || utils.NowUTC().Sub(*batches[0].CommittedAt) > time.Duration(r.cfg.GasOracleConfig.CheckCommittedBatchesWindowMinutes)*time.Minute, nil
+	// Also, we should check if it's a genesis batch. If so, skip the timeout check.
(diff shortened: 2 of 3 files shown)


colinlyguo (dismissed):
nice catch. a followup question: in normal cases, won't later batches be committed shortly after the genesis batch?

yiweichi:
> nice catch. a followup question: in normal cases, won't later batches be committed shortly after the genesis batch?

In scroll sdk, normally it takes a while to propose a batch, so normally there is only one batch(genesis) on database for a quite long time.

NazariiDenha (dismissed):

NazariiDenha (approved):

colinlyguo:
> > nice catch. a followup question: in normal cases, won't later batches be committed shortly after the genesis batch?
> 
> In scroll sdk, normally it takes a while to propose a batch, so normally there is only one batch(genesis) on database for a quite long time.

Would it be the same for batches 1, 2, and 3...? if so, can we increase the timeout to a longer period e.g. 1h?

Thegaram (approved):

colinlyguo (approved):

yiweichi:
> > > nice catch. a followup question: in normal cases, won't later batches be committed shortly after the genesis batch?
> > 
> > 
> > In scroll sdk, normally it takes a while to propose a batch, so normally there is only one batch(genesis) on database for a quite long time.
> 
> Would it be the same for batches 1, 2, and 3...? if so, can we increase the timeout to a longer period e.g. 1h?

If it's batch 1,2 and 3..., then I think it doesn't matter, because for those batch they actually have `CommittedAt` field. So there won't be `nil pointer dereference` error.

colinlyguo:
> > nice catch. a followup question: in normal cases, won't later batches be committed shortly after the genesis batch?
> 
> In scroll sdk, normally it takes a while to propose a batch, so normally there is only one batch(genesis) on database for a quite long time.

saw the nil pointer error in pr description. then lgtm.

PR #1502, opened 2024-08-26
fix: override config values use viper

yiweichi:

This PR aims to solve the bug when trying to override private key fields like `gas_oracle_sender_private_key` , `commit_sender_private_key`. 
For now, the way we override config fields is based on the tag of JSON marshaller, but the real tag of those private keys are `json:"-"`, so here I propose to handle it specifically as a private key when the tag is `-`.
Didn't use `Viper` because seems that will  involve quite amount of code changes.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label


georgehao:
Suggestion: use `viper` https://github.com/scroll-tech/scroll/pull/1492 to replace the current implementation

colinlyguo (approved):

colinlyguo:
> Suggestion: use `viper` #1492 to replace the current implementation

looks like the change is significant. @yiweichi what's the major part of code change? e.g. the self-implementing marshaling/unmarshaling private key part?

colinlyguo (dismissed):

yiweichi:
> > Suggestion: use `viper` #1492 to replace the current implementation
> 
> looks like the change is significant. @yiweichi what's the major part of code change? e.g. the self-implementing marshaling/unmarshaling private key part?

Yes, the major part would be marshaling/unmarshaling private key part.

colinlyguo:
> > > Suggestion: use `viper` #1492 to replace the current implementation
> > 
> > 
> > looks like the change is significant. @yiweichi what's the major part of code change? e.g. the self-implementing marshaling/unmarshaling private key part?
> 
> Yes, the major part would be marshaling/unmarshaling private key part.

added a pr to try it. still a bit messy though.

Thegaram (dismissed):

colinlyguo (approved):

georgehao (approved):

PR #1504, opened 2024-08-27
prover: upgrade to v0.13.1 / v0.12.2,   share params between versions.

lispc:
@amoylan2 i will still fixing some remaining testing.  You can first have a look to check the high level changes.

(the previous builder-of-handler-map pattern is quite complex dealing with lifetime.. since we alwasy just have 2 version, i simplify the codes.  wdyt? ) 

btw coordinator/internal/logic/verifier/verifier_test.go is not maintained.. each time i fix that file


georgehao (approved):

amoylan2:
I think it's a little bit complicate, I'm taking a try to make it simpler based on your change

roynalnaruto (approved):

amoylan2:
I submitted a new PR https://github.com/scroll-tech/scroll/pull/1507, could you have a look? @lispc

PR #1531, opened 2024-10-08
fix(sender): make sure gas price is above baseFee

0xmountaintop:

During the sdk test, we found that suggested gas price might be insufficent if the txpool have a few txs (cc @jonastheis any clue if you look into gas oracle logic?). and our txpool will reject a legacy tx if its gas price is below basefee.

Feedback from altlayer:
After running for a while, the suggested gas price will drop to 0.01GWei, and gas oracle sender couldn't send any tx successfully.
Then they have to manually send a tx with gas price = baseFee. After that, the gas oracle sender can resume working.

While I think the real solution is to fix the gas oracle, we can quick hack it in this way to make our sdk testnet working.

### PR title

- [ ] feat: A new feature
- [X] fix: A bug fix

### Deployment tag versioning

- [X] Yes

### Breaking change label

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
	- Updated version number to v4.4.78.
	- Enhanced gas price adjustment logic for legacy transaction resubmissions to ensure compliance with base fee requirements.

- **Bug Fixes**
	- Improved logging for transaction gas adjustments to provide clearer insights during resubmission.


rollup/internal/controller/sender/sender.go
```diff
@@ -357,6 +366,10 @@ func (s *Sender) resubmitTransaction(tx *gethTypes.Transaction, baseFee, blobBas
 		originalGasPrice := tx.GasPrice()
 		gasPrice := new(big.Int).Mul(originalGasPrice, escalateMultipleNum)
 		gasPrice = new(big.Int).Div(gasPrice, escalateMultipleDen)
+		baseFeeInt := new(big.Int).SetUint64(baseFee)
+		if gasPrice.Cmp(baseFeeInt) < 0 {
+			gasPrice = baseFeeInt
```
colinlyguo, line 367: may as well add a multiplier here? the drawback would be over-charging by the multiplier.

rollup/internal/controller/sender/sender.go
```diff
@@ -159,7 +159,16 @@ func (s *Sender) SendConfirmation(cfm *Confirmation) {
...
 	case LegacyTxType:
-		return s.estimateLegacyGas(target, data, fallbackGasLimit)
+		feeData, err := s.estimateLegacyGas(target, data, fallbackGasLimit)
+		if err != nil {
+			return nil, err
+		}
+		baseFeeInt := new(big.Int).SetUint64(baseFee)
+		maxGasPrice := new(big.Int).SetUint64(s.config.MaxGasPrice)
+		if feeData.gasPrice.Cmp(baseFeeInt) < 0 && baseFeeInt.Cmp(maxGasPrice) <= 0 {
+			feeData.gasPrice = baseFeeInt
```
colinlyguo: this is updated by `gasPrice, err := s.client.SuggestGasPrice(s.ctx)` in `estimateLegacyGas`.

colinlyguo (dismissed):
nice catch.

georgehao (dismissed):

rollup/internal/controller/sender/sender.go
0xmountaintop: `gasPrice` is already multiplied. if after being multiplied it's still below `baseFee` we should bump it. I prefer not to multiplying again here.

rollup/internal/controller/sender/sender.go
0xmountaintop: you mean 
```
func (s *EthereumAPI) GasPrice(ctx context.Context) (*hexutil.Big, error) {
	tipcap, err := s.b.SuggestGasTipCap(ctx)
	if err != nil {
		return nil, err
	}
	if head := s.b.CurrentHeader(); head.BaseFee != nil {
		tipcap.Add(tipcap, head.BaseFee)
	}
	return (*hexutil.Big)(tipcap), err
}
```
?


The weird thing is that we might still get suggeted gas price below basefee. See PR description and altlayer's feedback. 

TBH I haven't figured out why it will return gas price below basefee. We should investigate into it.

colinlyguo: maybe the block's base fee is updated after the txn is submitted.

rollup/internal/controller/sender/sender.go
colinlyguo: if the txn is submitted, and before it's included in a block, the block's base fee is updated by gas-oracle price update, the gas price will be < base fee again. same suspicion in another comment discussion.

rollup/internal/controller/sender/sender.go
0xmountaintop: reverted in 30a7b47b

colinlyguo (dismissed):

jonastheis:
> After running for a while, the suggested gas price will drop to 0.01GWei, and gas oracle sender couldn't send any tx successfully.

The question is where does this `0.01GWei` come from? Is it returned from the L2 node when querying the gas price or is it `s.config.MinGasTip`? See here: https://github.com/scroll-tech/scroll/blob/e3c87e6547b65233c0eeb07ce7411a7ff81d471d/rollup/internal/controller/sender/estimategas.go#L14

> Then they have to manually send a tx with gas price = baseFee. After that, the gas oracle sender can resume working.

I can only speculate but maybe something like this happened: the first transaction was not going through (send via `SendTransaction` [here](https://github.com/scroll-tech/scroll/blob/e3c87e6547b65233c0eeb07ce7411a7ff81d471d/rollup/internal/controller/relayer/l2_relayer.go#L340)) for example because the L2 base fee went up (due to L1 base fee changes). So this tx (and nonce) was "stuck". Now in the `resubmitTransaction` we didn't account for the base fee for legacy transactions which is why it couldn't recover.

colinlyguo:
> > After running for a while, the suggested gas price will drop to 0.01GWei, and gas oracle sender couldn't send any tx successfully.
> 
> The question is where does this `0.01GWei` come from? Is it returned from the L2 node when querying the gas price or is it `s.config.MinGasTip`? See here:
> 
> https://github.com/scroll-tech/scroll/blob/e3c87e6547b65233c0eeb07ce7411a7ff81d471d/rollup/internal/controller/sender/estimategas.go#L14
> 
> > Then they have to manually send a tx with gas price = baseFee. After that, the gas oracle sender can resume working.
> 
> I can only speculate but maybe something like this happened: the first transaction was not going through (send via `SendTransaction` [here](https://github.com/scroll-tech/scroll/blob/e3c87e6547b65233c0eeb07ce7411a7ff81d471d/rollup/internal/controller/relayer/l2_relayer.go#L340)) for example because the L2 base fee went up (due to L1 base fee changes). So this tx (and nonce) was "stuck". Now in the `resubmitTransaction` we didn't account for the base fee for legacy transactions which is why it couldn't recover.

agree. the probability of this speculation is higher when the network's L2 txns are mainly formed by gas oracle price update (txns updating L1 base fee and L1 blob base fee).

Thegaram (dismissed):

colinlyguo (approved):

colinlyguo (dismissed):

jonastheis (dismissed):

colinlyguo (approved):

georgehao (approved):

PR #1564, opened 2024-11-15
fix(sender): graceful restart

colinlyguo:
### Purpose or design rationale of this PR

This PR fixes a bug found during testing scroll-sdk.

Issue: the service restarts after sending the txn, and before adding the txn in db, so the txn's status will not be tracked or updated.
Fix: this PR first stores the txn in db, then sends the txn. if the txn fails to be sent, it will be covered and resubmitted by the resending logic.

### PR title

Your PR title must follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) (as we are doing squash merge for each PR), so it must start with one of the following [types](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type):

- [x] fix: A bug fix

### Deployment tag versioning

Has `tag` in `common/version.go` been updated or have you added `bump-version` label to this PR?

- [x] Yes

### Breaking change label

Does this PR have the `breaking-change` label?

- [x] No, this PR is not a breaking change

## Summary by CodeRabbit

- **New Features**
	- Updated version number to v4.4.75, indicating a new software release.
	- Enhanced transaction handling logic for improved clarity and error reporting.

- **Bug Fixes**
	- Improved error handling for transaction creation and sending processes.

- **Tests**
	- Refined test suite to align with updated transaction handling methods, ensuring robust validation of transaction processes.

PR #1567, opened 2024-11-28
fix(sender): graceful restart

colinlyguo:

This PR adds:
* db status rollback.
* blob tx fees precheck.

Note that:
* use permanent delete (Unscoped) instead of soft delete to prevent database bloat, as repeated SendTransaction failures could write a large number of transactions to the database.
* the db role should have "delete" permission after this PR.

### PR title

- [x] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

## Summary by CodeRabbit

- **New Features**
	- Updated version number to v4.4.79.
	- Introduced a method to delete pending transactions by transaction hash.

- **Bug Fixes**
	- Enhanced error handling and transaction management, including a rollback mechanism for failed transactions.

- **Improvements**
	- Added validation for gas fees in transaction creation to ensure they meet minimum thresholds.
	- Improved clarity and consistency in method naming for transaction status updates in tests.

### DIFF
--- a/rollup/internal/controller/sender/sender.go
+++ b/rollup/internal/controller/sender/sender.go
@@ -232,10 +232,10 @@ func (s *Sender) SendTransaction(contextID string, target *common.Address, data 
 	}
 
 	if err := s.client.SendTransaction(s.ctx, signedTx); err != nil {
-		// SendTransaction failed, mark the transaction as failed
-		if updateErr := s.pendingTransactionOrm.UpdatePendingTransactionStatusByTxHash(s.ctx, signedTx.Hash(), types.TxStatusSentFailed, nil); updateErr != nil {
-			log.Error("failed to mark transaction as sent failed", "tx hash", signedTx.Hash().String(), "from", s.transactionSigner.GetAddr().String(), "nonce", signedTx.Nonce(), "sendTxErr", err, "updateErr", updateErr)
-			return common.Hash{}, fmt.Errorf("failed to mark transaction as sent failed, err: %w", updateErr)
+		// Delete the transaction from the pending transaction table if it fails to send.
+		if updateErr := s.pendingTransactionOrm.DeletePendingTransactionByTxHash(s.ctx, signedTx.Hash()); updateErr != nil {
+			log.Error("failed to delete transaction", "tx hash", signedTx.Hash().String(), "from", s.transactionSigner.GetAddr().String(), "nonce", signedTx.Nonce(), "err", updateErr)
+			return common.Hash{}, fmt.Errorf("failed to delete transaction, err: %w", updateErr)
 		}
 
 		log.Error("failed to send tx", "tx hash", signedTx.Hash().String(), "from", s.transactionSigner.GetAddr().String(), "nonce", signedTx.Nonce(), "err", err)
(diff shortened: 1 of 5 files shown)


rollup/internal/controller/sender/sender.go
```diff
@@ -232,6 +232,12 @@ func (s *Sender) SendTransaction(contextID string, target *common.Address, data
 	}
 
 	if err := s.client.SendTransaction(s.ctx, signedTx); err != nil {
+		// Delete the transaction from the pending transaction table if it fails to send.
+		if updateErr := s.pendingTransactionOrm.DeletePendingTransactionByTxHash(s.ctx, signedTx.Hash()); updateErr != nil {
```
Thegaram: Why not move `s.pendingTransactionOrm.InsertPendingTransaction` after `s.client.SendTransaction` inside `SendTransaction`?

Also, what type of error do you have in mind here?

rollup/internal/controller/sender/sender.go
```diff
@@ -458,6 +464,15 @@ func (s *Sender) createReplacingTransaction(tx *gethTypes.Transaction, baseFee,
 				blobGasFeeCap = maxBlobGasPrice
 			}
 
+			// Check if any fee cap is less than double
+			doubledTipCap := new(big.Int).Mul(originalGasTipCap, big.NewInt(2))
+			doubledFeeCap := new(big.Int).Mul(originalGasFeeCap, big.NewInt(2))
+			doubledBlobFeeCap := new(big.Int).Mul(originalBlobGasFeeCap, big.NewInt(2))
+			if gasTipCap.Cmp(doubledTipCap) < 0 || gasFeeCap.Cmp(doubledFeeCap) < 0 || blobGasFeeCap.Cmp(doubledBlobFeeCap) < 0 {
```
Thegaram, line 471: Are we doubling in every gas bump? I thought that aggressive bumping only applies to blob fee cap

rollup/internal/controller/sender/sender.go
```diff
@@ -608,6 +623,23 @@ func (s *Sender) checkPendingTransaction() {
 			}
 
 			if err := s.client.SendTransaction(s.ctx, newSignedTx); err != nil {
+				// SendTransaction failed, need to rollback the previous database changes
```
Thegaram, line 626: Would it be clearer like this?

```go
if err := s.client.SendTransaction(s.ctx, newSignedTx); err != nil {
    tx.Rollback()
} else {
    tx.Commit()
}
```

Thegaram (dismissed):

rollup/internal/controller/sender/sender.go
colinlyguo: aggressive bumping only applies to blob tx, but blob pool will check if `gas fee cap`, `blob gas fee cap` and `gas tip` are all doubled in the meantime.

rollup/internal/controller/sender/sender.go
colinlyguo: this cannot prevent the service is forcefully shutdown by k8s right after sending the tx but not updating db. which is found by `altlayer`. then the sender will keep resubmitting and fail.

rollup/internal/controller/sender/sender.go
colinlyguo: > Also, what type of error do you have in mind here?

answered in another comment.

> move s.pendingTransactionOrm.InsertPendingTransaction after s.client.SendTransaction

yes. it's the previous implementation, while it cannot prevent forceful shutdown. but during graceful shutdown it does not have any problems (utilizing Golang's context), so this error never happened in sepolia and mainnet before.

rollup/internal/controller/sender/sender.go
colinlyguo: the fix pr of the bug found by `altlayer`'s testing: https://github.com/scroll-tech/scroll/pull/1564

jonastheis (dismissed):

georgehao (dismissed):

jonastheis (dismissed):

georgehao (approved):

PR #1571, opened 2024-12-11
perf: use strings.Builder

YESUNFEI:
In the ConvertBigIntArrayToString function, the big.Int is first converted to a string and then concatenated with commas. This can be optimized using strings.Builder, which avoids creating new string objects on each concatenation. The performance advantage is particularly significant when dealing with larger arrays.

### PR title

- [ ] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [ ] Yes

### Breaking change label

- [ ] No, this PR is not a breaking change
- [ ] Yes

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling with more descriptive messages for transaction data unpacking and calldata decoding.

- **Performance Improvements**
	- Optimized string concatenation process for better memory allocation and performance.

### DIFF
--- a/bridge-history-api/internal/utils/utils.go
+++ b/bridge-history-api/internal/utils/utils.go
@@ -173,6 +173,8 @@ func GetBlocksInRange(ctx context.Context, cli *ethclient.Client, start, end uin
 // ConvertBigIntArrayToString convert the big int array to string
 func ConvertBigIntArrayToString(array []*big.Int) string {
 	var sb strings.Builder
+	// Preallocate capacity (assuming ~20 chars per number plus separator)
+        sb.Grow(len(array) * 22)
 	for i, num := range array {
 		if i > 0 {
 			sb.WriteString(", ")
--- a/bridge-history-api/internal/utils/utils.go
+++ b/bridge-history-api/internal/utils/utils.go
@@ -173,8 +173,7 @@ func GetBlocksInRange(ctx context.Context, cli *ethclient.Client, start, end uin
 // ConvertBigIntArrayToString convert the big int array to string
 func ConvertBigIntArrayToString(array []*big.Int) string {
 	var sb strings.Builder
-	// Preallocate capacity (assuming ~20 chars per number plus separator)
-        sb.Grow(len(array) * 22)
+	sb.Grow(len(array) * 22)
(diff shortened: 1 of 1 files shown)


bridge-history-api/internal/utils/utils.go
```diff
@@ -172,13 +172,16 @@ func GetBlocksInRange(ctx context.Context, cli *ethclient.Client, start, end uin
 
 // ConvertBigIntArrayToString convert the big int array to string
 func ConvertBigIntArrayToString(array []*big.Int) string {
```
georgehao, line 174: can you add unit test for your changes?

FelixYe2023: I only modified the internal structure of the function and ran the `TestConvertBigIntArrayToString` test. The result passed successfully.

YESUNFEI: I only modified the internal structure of the function and ran the TestConvertBigIntArrayToString test. The result passed successfully.

colinlyguo: the code path running through this function (i.e. parsing `BatchWithdrawERC1155` and `BatchDepositERC1155`) is not the major path. thus it's likely that `ConvertBigIntArrayToString` won't affect the total perf.

bridge-history-api/internal/utils/utils.go
```diff
@@ -172,13 +172,15 @@ func GetBlocksInRange(ctx context.Context, cli *ethclient.Client, start, end uin
...
-	stringArray := make([]string, len(array))
+	var sb strings.Builder
+	sb.Grow(len(array) * 22)
```
colinlyguo, line 176: why use `22` here?

YESUNFEI: The number 22 is based on an estimated average length of the string representation of the big.Int numbers in the array.
A big.Int can have varying lengths depending on the size of the integer. However, for typical use cases, a 22-character length is likely a reasonable estimate to account for common values (including any digits and the potential minus sign for negative numbers). In the worst case, big.Int values can be much larger, but 22 is likely a safe upper bound for common integers.

PR #1572, opened 2024-12-11
opt: pre-allocate chunk blocks slice in chunk proposer

qcrao:

This PR improves performance by pre-allocating the capacity of chunk blocks slice based on the actual number of blocks to be processed.

What: Pre-allocate slice capacity in chunk proposer
Why: To reduce memory allocations and copying during chunk processing
How: Use len(blocks) as the initial capacity when creating the chunk.Blocks slice

### PR title

perf: pre-allocate chunk blocks slice in chunk proposer

- [x] perf: A code change that improves performance

### Deployment tag versioning

### Breaking change label

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling for chunk proposals, including checks for block limits.
	- New methods added for recording chunk metrics.

- **Bug Fixes**
	- Improved memory allocation for chunk blocks.

- **Chores**
	- Version updated from v4.4.81 to v4.4.83.

### DIFF
--- a/common/version/version.go
+++ b/common/version/version.go
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.82"
+var tag = "v4.4.83"
 
 var commit = func() string {
 	if info, ok := debug.ReadBuildInfo(); ok {
--- a/common/version/version.go
+++ b/common/version/version.go
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.81"
+var tag = "v4.4.82"
 
 var commit = func() string {
 	if info, ok := debug.ReadBuildInfo(); ok {
--- a/common/version/version.go
+++ b/common/version/version.go
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.84"
+var tag = "v4.4.83"
 
 var commit = func() string {
 	if info, ok := debug.ReadBuildInfo(); ok {
--- a/rollup/internal/controller/watcher/chunk_proposer.go
+++ b/rollup/internal/controller/watcher/chunk_proposer.go
@@ -295,6 +295,7 @@ func (p *ChunkProposer) proposeChunk() error {
 	}
 
 	var chunk encoding.Chunk
+	chunk.Blocks = make([]*encoding.Block, 0, len(blocks))
 	for i, block := range blocks {
 		chunk.Blocks = append(chunk.Blocks, block)
 


colinlyguo (dismissed):
lgtm. could you also add the patch version here? https://github.com/scroll-tech/scroll/blob/develop/common/version/version.go#L8

btw, `chunk.Blocks` array may be truncated (not fully used) in https://github.com/scroll-tech/scroll/pull/1572/files#diff-8d7032b59131e8543188def29571bf6769f1fd460cc04519c541cecdab5ef5f7R337, do you think pre-allocate will have downgraded performance related to it?

e.g. when there are many blocks in db, it will fetch 100 blocks each round, thus pre-allocating 100 blocks' memory, while a chunk usually contains much less than 100 blocks.

qcrao:
> lgtm. could you also add the patch version here? https://github.com/scroll-tech/scroll/blob/develop/common/version/version.go#L8
> 
> btw, `chunk.Blocks` array may be truncated in https://github.com/scroll-tech/scroll/pull/1572/files#diff-8d7032b59131e8543188def29571bf6769f1fd460cc04519c541cecdab5ef5f7R337, do you think pre-allocate will have downgraded performance related to it?
> 
> e.g. when there are many blocks in db, it will fetch 100 blocks each round, thus pre-allocating 100 blocks' memory, while a chunk usually contains much less than 100 blocks.

@colinlyguo  Good catch! In this case, pre-allocation still provides a net benefit:

1. The truncation operation (slice[:n-1]) doesn't involve memory reallocation or data copying - it just adjusts the slice length
2. Even with occasional truncation, having pre-allocated capacity helps avoid multiple grow-and-copy operations during the append operations
3. The memory for the pre-allocated array will be reused in subsequent operations rather than being reallocated

colinlyguo (dismissed):

colinlyguo:
> > lgtm. could you also add the patch version here? https://github.com/scroll-tech/scroll/blob/develop/common/version/version.go#L8
> > btw, `chunk.Blocks` array may be truncated in https://github.com/scroll-tech/scroll/pull/1572/files#diff-8d7032b59131e8543188def29571bf6769f1fd460cc04519c541cecdab5ef5f7R337, do you think pre-allocate will have downgraded performance related to it?
> > e.g. when there are many blocks in db, it will fetch 100 blocks each round, thus pre-allocating 100 blocks' memory, while a chunk usually contains much less than 100 blocks.
> 
> @colinlyguo Good catch! In this case, pre-allocation still provides a net benefit:
> 
> 1. The truncation operation (slice[:n-1]) doesn't involve memory reallocation or data copying - it just adjusts the slice length
> 2. Even with occasional truncation, having pre-allocated capacity helps avoid multiple grow-and-copy operations during the append operations
> 3. The memory for the pre-allocated array will be reused in subsequent operations rather than being reallocated

yes. you're right. it's "not fully used the pre-allocated memory" instead of "truncated".

As for point 3 I still have some doubts, would it be used again in next round? IMO the previous round's memory will be garbage collected.

georgehao (dismissed):

qcrao:
> > > lgtm. could you also add the patch version here? https://github.com/scroll-tech/scroll/blob/develop/common/version/version.go#L8
> > > btw, `chunk.Blocks` array may be truncated in https://github.com/scroll-tech/scroll/pull/1572/files#diff-8d7032b59131e8543188def29571bf6769f1fd460cc04519c541cecdab5ef5f7R337, do you think pre-allocate will have downgraded performance related to it?
> > > e.g. when there are many blocks in db, it will fetch 100 blocks each round, thus pre-allocating 100 blocks' memory, while a chunk usually contains much less than 100 blocks.
> > 
> > 
> > @colinlyguo Good catch! In this case, pre-allocation still provides a net benefit:
> > 
> > 1. The truncation operation (slice[:n-1]) doesn't involve memory reallocation or data copying - it just adjusts the slice length
> > 2. Even with occasional truncation, having pre-allocated capacity helps avoid multiple grow-and-copy operations during the append operations
> > 3. The memory for the pre-allocated array will be reused in subsequent operations rather than being reallocated
> 
> yes. you're right. it's "not fully used the pre-allocated memory" instead of "truncated".
> 
> As for point 3 I still have some doubts, would it be used again in next round? IMO the previous round's memory will be garbage collected.

Let me clarify: The pre-allocated memory from a previous round will indeed be subject to garbage collection once the chunk variable goes out of scope. Each new round will allocate fresh memory.

However, the pre-allocation is still beneficial because within each round:

It eliminates multiple grow-and-copy operations during the append phase
The "not fully used" space is just unused capacity, which has minimal overhead compared to the cost of dynamic resizing

georgehao (dismissed):

colinlyguo (dismissed):

colinlyguo (dismissed):

common/version/version.go
```diff
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.82"
+var tag = "v4.4.84"
```
colinlyguo:
```suggestion
var tag = "v4.4.83"
```

colinlyguo (approved):

georgehao (approved):

PR #1613, opened 2025-03-10
feat: openvm euclid v2

colinlyguo:

This PR:
* updates the openvm `euclidv2` dependencies of provers and coordinator's verifier.
* respects `euclidv2` in proof handling.
* adds `euclidv2` verifier in coordinator.
* adds `euclidv2` handler in prover.
* adds debug logs to print task details.
* supports `darwinv2`, `euclid`, `euclidv2` verifiers in coordinator.


common/types/message/message.go
```diff
@@ -39,20 +42,26 @@ const (
...
 
-// ChunkTaskDetail is a type containing ChunkTask detail.
-type ChunkTaskDetail struct {
+// LegacyChunkTaskDetail is a type containing ChunkTask detail.
+type LegacyChunkTaskDetail struct {
 	BlockHashes []common.Hash `json:"block_hashes"`
 }
 
+// EuclidV2ChunkTaskDetail is a type containing ChunkTask detail for euclid v2 chunk task.
+type EuclidV2ChunkTaskDetail struct {
```
noel2004: Maybe we do not need to separate ChunkTask for legacy (phase 1), for phase 1, prover can simply omit `prev_msg_queue` field while parsing the json.

common/types/message/message.go
```diff
@@ -39,20 +42,26 @@ const (
...
-	ChunkProofs   []ChunkProof `json:"chunk_proofs"`
-	BatchHeader   interface{}  `json:"batch_header"`
-	BlobBytes     []byte       `json:"blob_bytes"`
-	KzgProof      []byte       `json:"kzg_proof"`
-	KzgCommitment []byte       `json:"kzg_commitment"`
-	Challenge     common.Hash  `json:"challenge"`
+	ChunkInfos      []*ChunkInfo `json:"chunk_infos"`
+	ChunkProofs     []ChunkProof `json:"chunk_proofs"`
+	BatchHeader     interface{}  `json:"batch_header"`
+	BlobBytes       []byte       `json:"blob_bytes"`
```
noel2004, line 106: Can blob bytes being marshalled into base64 (prover expect base64 encoding here)

colinlyguo: It's a legacy field and can pass the batch task proving, so no problem in this field?

common/types/message/message.go
colinlyguo: OK. It's a suggestion from @yiweichi. Could you help check if `proving-sdk` can handle this case?

yiweichi: prover side can definitely handle this case with minor change. but phase 1 provers is already launched, I think we better to not upgrade them.

colinlyguo: OK. will keep it as it is.

colinlyguo: in phase-1 retest this works, so simplified the logic in this commit: https://github.com/scroll-tech/scroll/pull/1613/commits/e3332885edafb361e4815bdabcf9a93f7b7e00ba

common/libzkp/impl/src/verifier.rs
```diff
@@ -47,9 +49,19 @@ static mut VERIFIER_LOW: OnceCell<VerifierPair> = OnceCell::new();
 static mut VERIFIER_HIGH: OnceCell<VerifierPair> = OnceCell::new();
 
 pub fn init(config: VerifierConfig) {
-    let verifier = EuclidVerifier::new(&config.high_version_circuit.assets_path);
+    let verifier = EuclidVerifier::new(&config.low_version_circuit.assets_path);
```
colinlyguo: The path is the same as `euclidv2` verifier.

common/version/version.go
```diff
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.4.99"
+var tag = "v4.4.100"
```
georgehao: How about increasing the minor version

colinlyguo: good suggestion.

colinlyguo: changed in [dced0c6](https://github.com/scroll-tech/scroll/pull/1613/commits/dced0c6a821e757abdbb33486b5f330f84a148ec).

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -271,17 +288,13 @@ func (bp *BatchProverTask) getBatchTaskDetail(dbBatch *orm.Batch, chunkInfos []*
...
 	// Memory layout of `BlobDataProof`: used in Codec.BlobDataProofForPointEvaluation()
 	// | z       | y       | kzg_commitment | kzg_proof |
 	// |---------|---------|----------------|-----------|
 	// | bytes32 | bytes32 | bytes48        | bytes48   |
-	taskDetail.KzgProof = dbBatch.BlobDataProof[112:160]
-	taskDetail.KzgCommitment = dbBatch.BlobDataProof[64:112]
-	taskDetail.Challenge = common.Hash(dbBatch.BlobDataProof[0:32])
+	taskDetail.KzgProof = message.Byte48{Big: hexutil.Big(*new(big.Int).SetBytes(dbBatch.BlobDataProof[112:160]))}
+	taskDetail.KzgCommitment = message.Byte48{Big: hexutil.Big(*new(big.Int).SetBytes(dbBatch.BlobDataProof[64:112]))}
+
```
yiweichi, line 298: I suppose this coordinator will still need to support halo2 tasks for a while before the upgrade block. So I guess removing field `taskDetail.Challenge` and adding `taskDetail.ChallengeDigest` will brings about incompatibility issue?

colinlyguo: I think challenge is not used for `halo2` tasks. cc @roynalnaruto for the compatibility check.

yiweichi (approved):

coordinator/internal/logic/provertask/batch_prover_task.go
colinlyguo: Discussed with Rohit, currently, `Challenge` is not used in `DarwinV2`'s prover, so it indicates two things:
* Removing `Challenge' field is OK.
* Adding additional fields is OK.

noel2004: This should be the task defination for DarwinV2: https://github.com/scroll-tech/zkevm-circuits/blob/de04c6668125b732b47668d3bcd04723125762eb/prover/src/types.rs#L64

No "challenge" field occured

noel2004 (approved):

noel2004:
Just note: the new `challenge_digest` field is set in a new version of `da-codec`. We need to keep in mind that there is a new commit in the `main` branch of `da-codec` responses for this updating

I do not quite prefer such an implement (maybe at least we should branch/tag that updating) but anyway, acceptable for me

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -271,17 +288,13 @@ func (bp *BatchProverTask) getBatchTaskDetail(dbBatch *orm.Batch, chunkInfos []*
 	}
 	taskDetail.BatchHeader = batchHeader
 	taskDetail.BlobBytes = dbBatch.BlobBytes
-
-	if len(dbBatch.BlobDataProof) < 160 {
-		return nil, fmt.Errorf("blob data proof length is less than 160 bytes = %d, taskID: %s: %s", len(dbBatch.BlobDataProof), dbBatch.Hash, common.Bytes2Hex(dbBatch.BlobDataProof))
-	}
-
+	taskDetail.ChallengeDigest = common.HexToHash(dbBatch.ChallengeDigest)
```
noel2004, line 291: There is no clue for new developers that where these fields come from (in fact they are generated inside of `da-codec`), maybe we should add a comment about that

omerfirmak (approved):

PR #1637, opened 2025-03-26
Adapt to rc11 (unified phase)

noel2004:
This PR upgrade euclid dependings for rc11 with quite a few breaking changes:

+ The messages of ProvingTask need upgrade, a new fork_name field is required now

+ The deployment asset is updated, now there would be two bundle exe is packed in the released assets, one named "app.vmexe" and another named "app_euclidv1.vmexe", they have to be both deployed and loaded while creating the corresponding prover instance

### DIFF
--- a/zkvm-prover/src/zk_circuits_handler/euclid.rs
+++ b/zkvm-prover/src/zk_circuits_handler/euclid.rs
@@ -33,10 +33,7 @@ impl Phase {
         let dir_cache = Some(workspace_path.join("cache"));
         let path_app_exe = workspace_path.join("chunk/app.vmexe");
         let path_app_config = workspace_path.join("chunk/openvm.toml");
-        let segment_len = match self {
-            Phase::EuclidV1 => None,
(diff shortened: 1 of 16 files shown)


zkvm-prover/src/zk_circuits_handler/euclid.rs
```diff
@@ -6,35 +6,93 @@ use async_trait::async_trait;
...
+pub(crate) enum Phase {
+    EuclidV1,
+    EuclidV2,
+}
+
+impl Phase {
+    pub fn as_str(&self) -> &str {
+        match self {
+            Phase::EuclidV1 => "euclidv1",
+            Phase::EuclidV2 => "euclidv2",
```
colinlyguo, line 28:
```suggestion
            Phase::EuclidV1 => "euclid",
            Phase::EuclidV2 => "euclidV2",
```

colinlyguo: can test first.

PR #1652, opened 2025-05-08
refactor(coordinator): simplify logic post-Euclid

colinlyguo:

This PR removes halo2 prover logic and only keeps openvm prover logic.

### PR title

- [x] refactor: A code change that doesn't fix a bug, or add a feature, or improves performance

### Deployment tag versioning

- [x] Yes

### Breaking change label

## Summary by CodeRabbit

- **Refactor**
  - Removed legacy proof types and interfaces, consolidating all proof handling to use new OpenVM proof types for chunk, batch, and bundle proofs.
  - Simplified configuration and logic by removing support for "low version" circuits and mock mode throughout the system.
  - Updated APIs, controllers, and internal logic to exclusively support the new OpenVM proof structures and types.
  - Streamlined logging and verification key management to focus on the high version circuit and OpenVM verification keys.
  - Removed backward compatibility conditions and prover version special cases.
- **Chores**
  - Deleted all prover-related Rust source files, configuration files, build scripts, and CI workflows, effectively removing the prover component from the repository.
  - Cleaned up test and configuration files to reflect the removal of deprecated features and the prover component.
  - Removed scripts and tooling related to prover versioning and Rust toolchain pinning.
- **Tests**
  - Updated tests to use OpenVM proof types exclusively, removing references to deprecated proof types and adjusting assertions accordingly.
- **Style**
  - Reduced log verbosity and improved naming to reflect deprecated types where still referenced for compatibility.


common/types/message/message.go
```diff
@@ -99,22 +98,22 @@ func (e *Byte48) UnmarshalJSON(input []byte) error {
 // BatchTaskDetail is a type containing BatchTask detail.
 type BatchTaskDetail struct {
 	// use one of the string of EuclidFork / EuclidV2Fork
-	ForkName        string       `json:"fork_name"`
-	ChunkInfos      []*ChunkInfo `json:"chunk_infos"`
-	ChunkProofs     []ChunkProof `json:"chunk_proofs"`
```
Thegaram, line 104: Would it still make sense to maintain an abstract proof type? Or do you think it's unnecessary overhead?

coordinator/conf/config.json
```diff
@@ -7,17 +7,9 @@
     "batch_collection_time_sec": 180,
     "chunk_collection_time_sec": 180,
     "verifier": {
-      "mock_mode": true,
```
Thegaram, line 10: mock mode is not used anymore?

Thegaram (dismissed):

common/types/message/message.go
colinlyguo: Yeah this makes sense. Also thought of this for a while. I think we can always add it back to support multiple prover types, because they are just a few lines of code with two interfaces.

Currently I prefer to remove this abstract to reduce such type checks:
```
if openvmProof, ok := proof.(*message.OpenVMChunkProof); ok {
```
also it is more convenient when walking through the code. e.g., it jumps directly to the `openvm` types without mental effort.

coordinator/conf/config.json
colinlyguo: Yeah. iiuc this feature is never used in production, I guess that the original motivation is to bypass real verifier verification in some cases, while in all unit tests, "mock_verifier" tag is enabled, so coordinator uses mock verifier implementation so this path is never used.

coordinator/cmd/api/app/mock_app.go
```diff
@@ -90,17 +90,9 @@ func (c *CoordinatorApp) MockConfig(store bool) error {
 	cfg.ProverManager = &coordinatorConfig.ProverManager{
 		ProversPerSession: 1,
 		Verifier: &coordinatorConfig.VerifierConfig{
-			MockMode: true,
```
georgehao, line 93: why remove this line, the unit test don't need mock now?

colinlyguo: The unit test will enable "mock_verifier" by `-tags="mock_verifier"`, this code is for the real verifier and seems never to have been used.

georgehao (approved):

Thegaram (approved):

PR #1659, opened 2025-05-14
perf(relayer): add sequencer submission strategy with blob‐fee history and target price

ranchalp:

Implements a new sequencer submission strategy in the L2 relayer that decides when to publish batches based on recent blob‐fee conditions rather than immediately. It consists of:

- `fetchBlobFeeHistory(windowSec uint64)` – pulls the last `windowSec` seconds of blob‐fee samples from the `l1_block` table.

- `calculateTargetPrice(strat StrategyParams, firstTime time.Time, history []*big.Int)` – computes a dynamic “target” blob fee by applying either a percentile or EWMA baseline, then a relaxation curve (exponential or sigmoid) that increases toward the window deadline. Estimates per `windowSec` obtained from tests performed on traces.

- Integration into `ProcessPendingBatches` – before sending a commit transaction, the relayer compares the most recent blob fee to the target and only proceeds if the price is at or below the target, or if the configured time window has elapsed.

Resulting behavior: 

- Cost optimization: by delaying submission until blob fees are favorable, the protocol can reduce L1 data‐fee costs.

- Flexibility & safety: the relaxation curve ensures that batches aren’t starved indefinitely and will be submitted by the deadline even if prices remain high.

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  - Introduced a dynamic fee-submission strategy for batch processing, adapting submissions based on recent fee history and backlog size.
  - Added a method to retrieve counts of failed and pending batches to improve backlog monitoring.
  - Added querying of blob fee history over block ranges to support fee-based submission control.
- **Improvements**
  - Increased batch submission timeout from 5 minutes to 2 hours.
  - Added a configurable maximum backlog threshold to control batch submission flow.


ranchalp:
Will push new unit tests at `l2_relater_tests.go` tomorrow, but code can already be reviewed.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
 // ProcessPendingBatches processes the pending batches by sending commitBatch transactions to layer 1.
 func (r *Layer2Relayer) ProcessPendingBatches() {
 	// get pending batches from database in ascending order by their index.
-	dbBatches, err := r.batchOrm.GetFailedAndPendingBatches(r.ctx, r.cfg.BatchSubmission.MaxBatches)
+	allBatches, err := r.batchOrm.GetFailedAndPendingBatches(r.ctx, 0)
```
Thegaram: Better to limit at `r.cfg.BatchSubmission.MaxBatches`, otherwise we might need to read too many records.

If you just need it to check backlog size, you can add another count query for that.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
...
 	}
 
+	// if backlog outgrow max size, force‐submit enough oldest batches
+	backlogCount := len(allBatches)
+	backlogMax := 75 //r.cfg.BatchSubmission.BacklogMax
+
+	if len(allBatches) < r.cfg.BatchSubmission.MinBatches || len(allBatches) == 0 {
+		log.Debug("Not enough pending batches to submit", "count", len(allBatches), "minBatches", r.cfg.BatchSubmission.MinBatches, "maxBatches", r.cfg.BatchSubmission.MaxBatches)
+		return
+	}
```
Thegaram: The current batch submission trigger is "either collect `MinBatches` or timeout". Here you change it, what's the rationale for that? Is your point that with such a large window we should never have `<MinBatches` after timeout?

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
...
+
+	// return if not hitting target price
+	if backlogCount <= backlogMax {
+		windowSec := uint64(r.cfg.BatchSubmission.TimeoutSec)
+		strat, ok := bestParams[windowSec]
```
Thegaram: Can rename to `strategy` to make the linter happy.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1186,85 @@ func (r *Layer2Relayer) StopSenders() {
...
+		"number <= ?": latest,
+	}
+	recs, err := r.l1BlockOrm.GetL1Blocks(r.ctx, filters)
+	if err != nil {
+		return nil, fmt.Errorf("GetL1Blocks: %w", err)
+	}
+	hist := make([]*big.Int, len(recs))
+	for i, b := range recs {
+		hist[i] = new(big.Int).SetUint64(b.BlobBaseFee)
+	}
```
Thegaram: Better to just query the field (`BlobBaseFee`) and not the whole block.

Also, performance-wise this might be problematic (e.g. query 3600 blocks every couple of seconds). You might need to add a db index. Or better, maintain the fee history (or the moving average) in memory instead of querying blocks again every time.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
...
 	if err != nil {
 		log.Error("Failed to fetch pending L2 batches", "err", err)
 		return
 	}
 
+	// if backlog outgrow max size, force‐submit enough oldest batches
+	backlogCount := len(allBatches)
+	backlogMax := 75 //r.cfg.BatchSubmission.BacklogMax
+
+	if len(allBatches) < r.cfg.BatchSubmission.MinBatches || len(allBatches) == 0 {
```
jonastheis:
```suggestion
	if len(allBatches) < r.cfg.BatchSubmission.MinBatches {
```
I think the second check is not necessary. Also with Peter's suggestion to do a count query you can do this check at the very beginning (and keep the `backlogMax` stuff below it for better separation and readability.

rollup/internal/controller/relayer/l2_relayer.go
jonastheis: My understanding is: we don't necessarily want to commit even if we reach `MinBatches`. We still want to wait and commit at the best time. Or after a timeout we force the submission. The timeout is variable though, influenced by the timeout window (max) and backlog of batches. Worst case, we commit at an expensive price towards the end of the window.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1186,85 @@ func (r *Layer2Relayer) StopSenders() {
...
+		baseline = ewma
+	default:
+		baseline = data[n-1]
+	}
+	// relaxation
+	age := time.Since(firstTime).Seconds()
+	frac := age / float64(windowSec)
+	var adjusted float64
+	switch strat.RelaxType {
+	case "exponential":
```
jonastheis: better to use constants for this. can be either string constants or better with `iota` modifier a Go style enum

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
...
+	backlogMax := 75 //r.cfg.BatchSubmission.BacklogMax
+
+	if len(allBatches) < r.cfg.BatchSubmission.MinBatches || len(allBatches) == 0 {
+		log.Debug("Not enough pending batches to submit", "count", len(allBatches), "minBatches", r.cfg.BatchSubmission.MinBatches, "maxBatches", r.cfg.BatchSubmission.MaxBatches)
+		return
+	}
+
+	// return if not hitting target price
+	if backlogCount <= backlogMax {
+		windowSec := uint64(r.cfg.BatchSubmission.TimeoutSec)
```
jonastheis: Since this parameter can only change by restart, I think it would be better to assign the strategy in the constructor instead of this clunky few lines being here. Either startup can fail or set a reasonable default like you do here.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -268,12 +287,59 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
...
+	backlogCount := len(allBatches)
+	backlogMax := 75 //r.cfg.BatchSubmission.BacklogMax
+
+	if len(allBatches) < r.cfg.BatchSubmission.MinBatches || len(allBatches) == 0 {
+		log.Debug("Not enough pending batches to submit", "count", len(allBatches), "minBatches", r.cfg.BatchSubmission.MinBatches, "maxBatches", r.cfg.BatchSubmission.MaxBatches)
+		return
+	}
+
+	// return if not hitting target price
+	if backlogCount <= backlogMax {
```
jonastheis: somehow this code block is very ugly. Maybe you can put the logic into a function and then just return with the error message that you want to log. In that case we can continue and submit immediately (and log the error message). this way we can use the Go typical guard style

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1186,85 @@ func (r *Layer2Relayer) StopSenders() {
...
+	// convert to float64 Gwei
+	data := make([]float64, n)
+	for i, v := range history {
+		f, _ := new(big.Float).Quo(new(big.Float).SetInt(v), big.NewFloat(1e9)).Float64()
+		data[i] = f
+	}
+	var baseline float64
+	switch strat.BaselineType {
+	case "pct_min":
+		sort.Float64s(data)
```
jonastheis: This is probably also relatively expensive after the repeated DB query. Like Peter said would be better to keep this in memory.

Maybe something like this construction https://github.com/scroll-tech/go-ethereum/blob/e62c6f08e2cb233ca0e14d01bc3f41f2edaea369/rollup/da_syncer/batch_queue.go#L20
with a sorted heap and a shrinkingmap in memory could work.

Every time you check you would only need to verify which min/max blocknumbers you expect to be in the map. if they are not there then you query them from DB and add them (this way also after startup should be handled automatically). if an entry is too old you delete it out of the heap and map.

jonastheis (changes requested):

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: I removed this to keep the original semantic of "either collect MinBatches or timeout" (adding extra conditions if not in the timeout on top of having collected enough batches.

rollup/internal/orm/batch.go
```diff
@@ -218,18 +218,32 @@ func (o *Batch) GetRollupStatusByHashList(ctx context.Context, hashes []string)
...
+	}
+	return count, nil
+}
+
 // GetFailedAndPendingBatches retrieves batches with failed or pending status up to the specified limit.
 // The returned batches are sorted in ascending order by their index.
 func (o *Batch) GetFailedAndPendingBatches(ctx context.Context, limit int) ([]*Batch, error) {
-	if limit <= 0 {
-		return nil, errors.New("limit must be greater than zero")
+	if limit < 0 {
```
jonastheis: Revert these changes if not necessary anymore

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -274,6 +341,23 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
...
 
+	// if backlog outgrow max size, force‐submit enough oldest batches
+	backlogCount, err := r.batchOrm.GetFailedAndPendingBatchesCount(r.ctx)
+	if err != nil {
+		log.Error("Failed to fetch pending L2 batches", "err", err)
+		return
+	}
+
+	// return if not hitting target price
+	if backlogCount <= r.cfg.BatchSubmission.BacklogMax {
```
jonastheis: Now this function is much easier to read and reason about. One thing that we might consider but is probably okay for now: we always do the calculation, even if we don't have at least `MinBatches`. Only to then afterwards exit bc we don't have enough batches.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -274,6 +341,23 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
...
+		oldest := dbBatches[0].CreatedAt
+		if skip, msg := r.skipSubmitByFee(oldest); skip {
+			log.Debug(msg)
+			return
+		}
+		// if !skip, we fall through and submit immediately
```
jonastheis: Now we don't log the reason of `!skip`. Instead of a msg as a string I think it's better to use an error type and just extract the message

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1204,116 @@ func (r *Layer2Relayer) StopSenders() {
...
+			return nil, fmt.Errorf("GetBlobFeesInRange: %w", err)
+		}
+		// append them
+		for _, v := range raw {
+			r.feeHistory = append(r.feeHistory, new(big.Int).SetUint64(v))
+			r.lastFetchedBlock++
+		}
+	}
+
+	maxLen := int(windowSec / 12)
```
jonastheis: use a constant for 12

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -141,6 +189,25 @@ func NewLayer2Relayer(ctx context.Context, l2Client *ethclient.Client, db *gorm.
...
+	strategy, ok := bestParams[windowSec]
+	if !ok {
+		return nil, fmt.Errorf(
+			"unsupported BatchSubmission.TimeoutSec: %d (must be one of %v)",
+			windowSec, maps.Keys(bestParams),
+		)
+	}
+	layer2Relayer.batchStrategy = strategy
+
+	latest, err := layer2Relayer.l1BlockOrm.GetLatestL1BlockHeight(ctx)
```
jonastheis: Instead of having duplicate code can't we simply make `fetchBlobFeeHistory` recover itself if there's no previous state?

As far as I can see we only need to check if `layer2Relayer.lastFetchedBlock == 0` and then set it accordinlgy.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1204,116 @@ func (r *Layer2Relayer) StopSenders() {
...
+		}
+	}
+
+	maxLen := int(windowSec / 12)
+	if len(r.feeHistory) > maxLen {
+		r.feeHistory = r.feeHistory[len(r.feeHistory)-maxLen:]
+	}
+	// return a copy
+	out := make([]*big.Int, len(r.feeHistory))
+	copy(out, r.feeHistory)
```
jonastheis: Why copy here?

ranchalp: You are right, this is not needed. I thought because of the sort in calculate target price when using PctMin but the sort already takes a different slice.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Fair enough. Done.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Good point. I changed the code to do this, but something that I did while doing that is only setting forceSubmit to true if the oldest batch by index must be force submitted, unlike now that this is check for all the batches in dbBatches. I argue this is ok as the smallest index must be the one that must be submitted next. What do you think?

rollup/internal/orm/batch.go
ranchalp: done.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Done.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -189,23 +190,7 @@ func NewLayer2Relayer(ctx context.Context, l2Client *ethclient.Client, db *gorm.
...
-	}
-	layer2Relayer.batchStrategy = strategy
-
-	latest, err := layer2Relayer.l1BlockOrm.GetLatestL1BlockHeight(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get latest L1 block height: %v", err)
-	}
-	layer2Relayer.lastFetchedBlock = latest - uint64(layer2Relayer.cfg.BatchSubmission.TimeoutSec)/12 // start ~window seconds ago
-	if _, err = layer2Relayer.fetchBlobFeeHistory(uint64(layer2Relayer.cfg.BatchSubmission.TimeoutSec)); err != nil {
+	if _, err := layer2Relayer.fetchBlobFeeHistory(uint64(layer2Relayer.cfg.BatchSubmission.TimeoutSec)); err != nil {
```
jonastheis: This is exactly called as is in `skipSubmitByFee`. I think we can just remove it from the constructor no?

rollup/internal/controller/relayer/l2_relayer.go
jonastheis: I think this is okay

jonastheis (commented):
Would be good to add a high level comment to `ProcessPendingBatches`: 

We submit batches if:
- the first batch is too old -> forceSubmit
- backlogCount > r.cfg.BatchSubmission.BacklogMax -> forceSubmit
- we have at least min count AND price is good

I committed a small refactor as I wanted to try it locally first

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1120,6 +1199,119 @@ func (r *Layer2Relayer) StopSenders() {
...
+}
+
+// calculateTargetPrice applies pct_min/ewma + relaxation to get a BigInt target
+func calculateTargetPrice(windowSec uint64, strategy StrategyParams, firstTime time.Time, history []*big.Int) *big.Int {
+	n := len(history)
+	if n == 0 {
+		return big.NewInt(0)
+	}
+	// convert to float64 Gwei
+	data := make([]float64, n)
```
jonastheis: why do we copy from `[]big.Int` to `[]float64` here? Can't we run into precision and rounding issues then? Shouldn't it be possible to do all operations in `big.Int`? 

Also as described in [this comment](https://github.com/scroll-tech/scroll/pull/1659/files#r2091051562) we still sort the data array every time, which might be quite expensive. 

I think this is okay for now, we should address this in the next version though.

jonastheis:
I'm okay with merging this once all comments except https://github.com/scroll-tech/scroll/pull/1659#discussion_r2094716744 are resolved. 

Other things we should consider in the next version:
1.  extracting all the fee related functionality into it's own struct to encapsulate the complexity better. The only thing that object needs to expose is the `skipSubmitByFee` function. Everything else can be abstracted/hidden away from the process of processing pending batches. 
2. unit tests for this new component to test the fee mechanism and its complexity in isolation
3. integration tests (the usual l2relayer tests) that incorporate some tests with specific L1 fee scenarios

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Yes, it can be removed. It would just prepare the big chunk of the history on construction instead of during the loop. Otherwise it is fine.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: I thought the mention was to keep in memory the history, not the call to sort. The sort operation should be of negligible performance impact (we are talking about 7200 elements to sort through for a 24 hour period). Do you think it makes sense to invest time in removing this? 

Similar argument goes with float64. Dealing with big.Int will make for a more convoluted code when doing the exponent for the exponential and sigmoid relaxations. The loss of precision would be negligible (a couple wei), unless the blob base fee grows beyond unless the blob base fee grows to the order of millions of Gwei (i.e. ≳10⁶ Gwei, or ≳10¹⁵ Wei), at which point you’d still only lose a handful of Wei to floating-point imprecision in comparison. 

I have thus for the moment limitted the conversion to float to the end of the calculation of the baseline price, but I can still change all to big.Int all around if desired (probably in a new PR not to block this one). Same goes for the sort function.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: You are right, done.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Done, thanks.

rollup/internal/controller/relayer/l2_relayer.go
ranchalp: Thanks, done.

jonastheis (dismissed):

jonastheis (approved):

Thegaram (approved):

PR #1693, opened 2025-07-07
feat(rollup-relayer): support Validium

colinlyguo:

Support Validium in rollup-relayer.

### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [x] Yes

### Breaking change label

## Summary by CodeRabbit

* **New Features**
  * Added support for a new "Validium" mode, enabling relayer operation with Validium smart contracts.
  * Introduced configuration option to enable or disable Validium mode.
  * Enhanced batch processing, payload construction, and metrics calculation to support Validium mode.
  * Integrated new ABI for interacting with the Validium contract on ScrollChain.

* **Chores**
  * Updated version to v4.5.29.


rollup/abi/validium_abi.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+var (
+	// ValidiumABI holds information about Validium's context and available invokable methods.
+	ValidiumABI *abi.ABI
+)
+
+func init() {
+	ValidiumABI, _ = ValidiumMetaData.GetAbi()
+}
```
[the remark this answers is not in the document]
colinlyguo: This is a valid suggestion, but I will keep it as it is to align with code style in `bridge_abi.go`.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -982,6 +1057,22 @@ func (r *Layer2Relayer) constructFinalizeBundlePayloadCodecV7(dbBatch *orm.Batch
...
+		"finalizeBundle",
+		dbBatch.BatchHeader,
+		new(big.Int).SetUint64(endChunk.TotalL1MessagesPoppedBefore+endChunk.TotalL1MessagesPoppedInChunk),
+		aggProof.Proof(),
+	)
+	if packErr != nil {
+		return nil, fmt.Errorf("failed to pack finalizeBundlePostEuclidV2NoProof: %w", packErr)
+	}
+	return calldata, nil
+}
```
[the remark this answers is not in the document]
colinlyguo: Addressed in [c8ab62e](https://github.com/scroll-tech/scroll/pull/1693/commits/c8ab62ed9a2004f9ab690c9f69117183f7db3373).

rollup/internal/utils/utils.go
```diff
@@ -119,8 +124,28 @@ type BatchMetadata struct {
...
+	var commitment common.Hash
+	if len(b.Blocks) > 0 {
+		lastBlock := b.Blocks[len(b.Blocks)-1]
+		commitment = lastBlock.Header.Hash()
+	}
+	copy(batchBytes[105:137], commitment[0:32]) // data commitment
+
+	hash := crypto.Keccak256Hash(batchBytes)
+	return batchBytes, hash
+}
```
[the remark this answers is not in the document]
colinlyguo: Addressed in [c8ab62e](https://github.com/scroll-tech/scroll/pull/1693/commits/c8ab62ed9a2004f9ab690c9f69117183f7db3373).

rollup/abi/validium_abi.go
```diff
@@ -0,0 +1,22 @@
...
+
+// Generated manually from abigen.
+
+// ValidiumMetaData contains all meta data concerning the ScrollChain contract.
```
Thegaram:
```suggestion
// ValidiumMetaData contains all meta data concerning the ScrollChainValidium contract.
```

rollup/abi/validium_abi.go
```diff
@@ -0,0 +1,22 @@
...
+var ValidiumMetaData = &bind.MetaData{
```
Thegaram, line 20: Can you please generate these from the latest contract version? (Ignore this comment if you already did it.)

rollup/cmd/rollup_relayer/app/app.go
```diff
@@ -108,6 +108,7 @@ func action(ctx *cli.Context) error {
 
 	chunkProposer := watcher.NewChunkProposer(subCtx, cfg.L2Config.ChunkProposerConfig, minCodecVersion, genesis.Config, db, registry)
 	batchProposer := watcher.NewBatchProposer(subCtx, cfg.L2Config.BatchProposerConfig, minCodecVersion, genesis.Config, db, registry)
+	batchProposer.SetValidiumMode(cfg.L2Config.RelayerConfig.ValidiumMode)
```
Thegaram: Why not just pass it in `NewBatchProposer`?

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -466,10 +481,18 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 	codecVersion := encoding.CodecVersion(firstBatch.CodecVersion)
 	switch codecVersion {
 	case encoding.CodecV7, encoding.CodecV8:
-		calldata, blobs, maxBlockHeight, totalGasUsed, err = r.constructCommitBatchPayloadCodecV7(batchesToSubmit, firstBatch, lastBatch)
-		if err != nil {
-			log.Error("failed to construct constructCommitBatchPayloadCodecV7 payload for V7", "codecVersion", codecVersion, "start index", firstBatch.Index, "end index", lastBatch.Index, "err", err)
-			return
+		if r.cfg.ValidiumMode {
+			calldata, maxBlockHeight, totalGasUsed, err = r.constructCommitBatchPayloadValidium(batchesToSubmit, firstBatch)
```
Thegaram: This made me think. If we have multiple pending batches, maybe we should actually recreate them into a single batch; there's no point in committing in multiple steps (apart from controlling the prover workload). Something to consider in a later version.

rollup/internal/utils/utils.go
```diff
@@ -119,8 +124,57 @@ type BatchMetadata struct {
...
+	}
+
+	// Batch header field sizes
+	const (
+		versionSize      = 1
+		indexSize        = 8
+		parentHashSize   = 32
+		stateRootSize    = 32
+		withdrawRootSize = 32
+		commitmentSize   = 32
```
Thegaram: We might use larger commitment in the future; for now let's keep it as 32 bytes.

rollup/internal/utils/utils.go
```diff
@@ -119,8 +124,57 @@ type BatchMetadata struct {
...
+
+	batchBytes[versionOffset] = uint8(codecVersion)                                                                        // version
+	binary.BigEndian.PutUint64(batchBytes[indexOffset:indexOffset+indexSize], b.Index)                                     // batch index
+	copy(batchBytes[parentHashOffset:parentHashOffset+parentHashSize], b.ParentBatchHash[0:parentHashSize])                // parentBatchHash
+	copy(batchBytes[stateRootOffset:stateRootOffset+stateRootSize], b.StateRoot().Bytes()[0:stateRootSize])                // postStateRoot
+	copy(batchBytes[withdrawRootOffset:withdrawRootOffset+withdrawRootSize], b.WithdrawRoot().Bytes()[0:withdrawRootSize]) // postWithdrawRoot
+
+	// For validium mode, use the last block hash as commitment to the off-chain data
+	var commitment common.Hash
+	if len(b.Blocks) > 0 {
```
Thegaram: We should not allow the case when `len(blocks) == 0`.

Thegaram (dismissed):

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: I think currently it's OK because timeout seems to be the only limit of batch proposing, so that the batch can be very large. It would be rare that there are many pending batches in db unless batch commit is stuck. I may misunderstand the real issue and miss some cases.

To apply batch merging, since inter-component coordination, such as freeze `batch-proposer` and `bundle-proposer`, and tweak db, would be a bit tricky.

Another way is to support `commitBatches` in the contract (though this only partially solved the cost problem).

colinlyguo: Or do these limits of v7 and v8 chunk/batch still apply?
```
    "chunk_proposer_config": {
      "propose_interval_milliseconds": 100,
      "max_block_num_per_chunk": 100,
      "max_l2_gas_per_chunk": 20000000,
      "chunk_timeout_sec": 300,
      "max_uncompressed_batch_bytes_size": 4194304
    },
    "batch_proposer_config": {
      "propose_interval_milliseconds": 1000,
      "batch_timeout_sec": 300,
      "max_chunks_per_batch": 45,
      "max_uncompressed_batch_bytes_size": 4194304
    },
    "bundle_proposer_config": {
      "max_batch_num_per_bundle": 20,
      "bundle_timeout_sec": 36000
    },
```

rollup/abi/validium_abi.go
colinlyguo: Updated with this commit: [f4d6e67](https://github.com/scroll-tech/scroll/pull/1693/commits/f4d6e67c90a6a555f1bb614fda2d53bb23aebfaf). We can update it again after contract implementation is finalized (maybe merged).

rollup/cmd/rollup_relayer/app/app.go
colinlyguo: This is added to not modify the parameters of `NewBatchProposer`, but true based on your comment it's clearer, so changed to pass it in `NewBatchProposer` with this commit: [f4d6e67](https://github.com/scroll-tech/scroll/pull/1693/commits/f4d6e67c90a6a555f1bb614fda2d53bb23aebfaf).

rollup/internal/utils/utils.go
colinlyguo: Added a comment in this commit: [f4d6e67](https://github.com/scroll-tech/scroll/pull/1693/commits/f4d6e67c90a6a555f1bb614fda2d53bb23aebfaf), also other TODO comments in related `commitment` places.

rollup/internal/utils/utils.go
colinlyguo: Added a check and error return in [f4d6e67](https://github.com/scroll-tech/scroll/pull/1693/commits/f4d6e67c90a6a555f1bb614fda2d53bb23aebfaf).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Also found that currently it will not commit one batch at a time, fixed in this commit: [7dfb8b5](https://github.com/scroll-tech/scroll/pull/1693/commits/7dfb8b5aee658ed0399590e7dea7d4954dac81da).

zimpha (dismissed):

rollup/internal/controller/relayer/l2_relayer.go
Thegaram: For now let's keep the existing limits (we can adjust via config anyway).

colinlyguo: That makes sense, too. I also noticed that the blob size limit is not easily walk-around because it is embedded in error paths of da-codec functions, such as `NewDABatch` or `CheckCompatibility`.

rollup/cmd/rollup_relayer/app/app.go
```diff
@@ -96,6 +96,13 @@ func action(ctx *cli.Context) error {
 		log.Crit("cfg.L2Config.ChunkProposerConfig.MaxL2GasPerChunk must be greater than 0")
 	}
 
+	if cfg.L2Config.RelayerConfig.ValidiumMode {
+		// Force single batch submission for validium mode, this looks a bit hacky but it avoids more changes in the relayer code.
```
georgehao: a better change is set the default value as 1, when set it in config, then use the config value

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -466,10 +482,22 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
...
+			}
+			calldata, maxBlockHeight, totalGasUsed, err = r.constructCommitBatchPayloadValidium(batchesToSubmit[0])
+			if err != nil {
+				log.Error("failed to construct validium payload", "codecVersion", codecVersion, "index", batchesToSubmit[0].Batch.Index, "err", err)
+				return
+			}
+		} else {
+			calldata, blobs, maxBlockHeight, totalGasUsed, err = r.constructCommitBatchPayloadCodecV7(batchesToSubmit, firstBatch, lastBatch)
+			if err != nil {
+				log.Error("failed to construct normal payload", "codecVersion", codecVersion, "start index", firstBatch.Index, "end index", lastBatch.Index, "err", err)
```
georgehao, line 501: I don't like this code style. How about passing the `ValidiumMode` to da-codec, making the `constructCommitBatchPayloadValidium` and `constructCommitBatchPayloadCodecV7` as the same function

rollup/cmd/rollup_relayer/app/app.go
colinlyguo: I agree that setting a default value would be cleaner. However, even if we remove this override, we should add a validity check to ensure batches are contiguous since the contract doesn't verify batch gaps and incorrect commits would be difficult to handle. wdyt?

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: The two functions call different ABI interfaces - `commitBatch` vs `commitBatches` - so this distinction is better handled at the rollup-relayer.

rollup/cmd/rollup_relayer/app/app.go
colinlyguo: Refactored the logic in [72aa00e](https://github.com/scroll-tech/scroll/pull/1693/commits/72aa00ee5072450239869c4d10e777c67db18465).

Thegaram (dismissed):

georgehao (dismissed):

PR #1714, opened 2025-07-29
feat(rollup-relayer): add sanity checks before committing and finalizing

colinlyguo:
### PR title

- [x] feat: A new feature

### Deployment tag versioning

- [x] Yes

### Breaking change label

## Summary by CodeRabbit

* **New Features**
  * Added extensive data validation and sanity checks to enhance the integrity of batch and chunk processing.
  * Introduced a method to retrieve batch records by their hash.

* **Bug Fixes**
  * Corrected parent batch hash and block number assignments in test setups for improved accuracy.

* **Tests**
  * Updated tests to align with new validation logic and ensure accurate data relationships.

* **Chores**
  * Updated version tag to v4.5.39.


rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1242,3 +1394,258 @@ func addrFromSignerConfig(config *config.SignerConfig) (common.Address, error) {
...
+	} else {
+		// For the first batch, check continuity with the parent batch from database
+		parentBatch, err := r.batchOrm.GetBatchByHash(r.ctx, batch.Batch.ParentBatchHash)
+		if err != nil {
+			return fmt.Errorf("failed to get parent batch %s for batch %d: %w", batch.Batch.ParentBatchHash, batch.Batch.Index, err)
+		}
+		if batch.Batch.Index != parentBatch.Index+1 {
+			return fmt.Errorf("first batch index is not sequential with parent: parent batch index %d, current batch index %d", parentBatch.Index, batch.Batch.Index)
+		}
+	}
```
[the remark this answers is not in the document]
colinlyguo: No. If it's not found, just error out. The genesis batch will be filled in another place.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -987,17 +1071,42 @@ func (r *Layer2Relayer) constructCommitBatchPayloadCodecV7(batchesToSubmit []*db
...
+
+	if len(calldata) == 0 {
+		return nil, nil, 0, 0, fmt.Errorf("generated calldata is empty")
+	}
+
 	return calldata, blobs, maxBlockHeight, totalGasUsed, nil
 }
 
 func (r *Layer2Relayer) constructCommitBatchPayloadValidium(batch *dbBatchWithChunks) ([]byte, uint64, uint64, error) {
+	// Basic sanity checks
```
georgehao: all the basic can unify to a more common function: `func sanityCheck(...)`

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -473,6 +491,12 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		log.Info("Forcing submission of batches due to timeout", "batch index", batchesToSubmit[0].Batch.Index, "first block created at", oldestBlockTimestamp)
 	}
 
+	// Sanity checks before constructing the transaction
+	if err := r.sanityChecksBeforeConstructingTransaction(batchesToSubmit); err != nil {
+		log.Error("Sanity checks failed before constructing transaction", "batches count", len(batchesToSubmit), "first batch index", batchesToSubmit[0].Batch.Index, "last batch index", batchesToSubmit[len(batchesToSubmit)-1].Batch.Index, "err", err)
```
georgehao: do we need to add a metric when get the sanityCheck failed error.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1242,3 +1423,263 @@ func addrFromSignerConfig(config *config.SignerConfig) (common.Address, error) {
...
+
+	// If there are any L1 messages processed up to this batch, post hash should not be zero
+	totalL1MessagesProcessed := lastChunk.TotalL1MessagesPoppedBefore + lastChunk.TotalL1MessagesPoppedInChunk
+	if totalL1MessagesProcessed > 0 && postL1MsgQueueHash == (common.Hash{}) {
+		return fmt.Errorf("batch %d post L1 message queue hash is zero but %d L1 messages were processed in total", batch.Batch.Index, totalL1MessagesProcessed)
+	}
+
+	// Prev and post queue hashes should be different if L1 messages were processed in this batch
+	if batchTotalL1MessagesInBatch > 0 && prevL1MsgQueueHash == postL1MsgQueueHash {
+		return fmt.Errorf("batch %d has same prev and post L1 message queue hashes but processed %d L1 messages in this batch", batch.Batch.Index, batchTotalL1MessagesInBatch)
```
georgehao: L1049-1062 is same with this section, can we reuse `validateMessageQueueConsistency` for L1049-1062

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1242,3 +1423,263 @@ func addrFromSignerConfig(config *config.SignerConfig) (common.Address, error) {
...
+	}
+
+	return prevChunk, nil
+}
+
+// validateBatchesDetailed performs detailed validation on each batch
+func (r *Layer2Relayer) validateBatchesDetailed(batchesToSubmit []*dbBatchWithChunks, prevChunkFromPrevBatch *orm.Chunk) error {
+	for i, batch := range batchesToSubmit {
+		if err := r.validateSingleBatch(batch, i, batchesToSubmit, prevChunkFromPrevBatch); err != nil {
+			return err
```
georgehao: add a log here

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1242,3 +1423,263 @@ func addrFromSignerConfig(config *config.SignerConfig) (common.Address, error) {
 		return common.Address{}, fmt.Errorf("failed to determine signer address, unknown signer type: %v", config.SignerType)
 	}
 }
+
+// sanityChecksBeforeConstructingTransaction performs sanity checks before constructing a transaction.
+func (r *Layer2Relayer) sanityChecksBeforeConstructingTransaction(batchesToSubmit []*dbBatchWithChunks) error {
```
georgehao: I suggest splitting the sanity checks to an independent file: `l2_relayer_santiy.go`, currently this file `l2_relayer.go` is too long

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: I think it's OK since there are many alerts about committing/finalizing delay.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: I think it's OK because in the first layer of error propagation, the log is very detailed. So it will be printed in the place that catches this error.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: They are case by case, depending on the function's parameters, so we can simply keep them inside the function. The unified sanity checks are validated in `sanityChecksBeforeConstructingTransaction`.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Makes sense. Moved the implementation to `l2_relayer_santiy.go in commit`: [604e33a](https://github.com/scroll-tech/scroll/pull/1714/commits/604e33ae7b2da98eef7356d98ae7f8f0add2720c).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Sounds good. Unified in commit: [c67888c](https://github.com/scroll-tech/scroll/pull/1714/commits/c67888c0bb860286d60a09d619481ae4849cf638)

yiweichi (dismissed):

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -977,6 +1030,16 @@ func (r *Layer2Relayer) constructCommitBatchPayloadCodecV7(batchesToSubmit []*db
 			Blocks:                 batchBlocks,
 		}
```
jonastheis: `len(batchBlocks) == 0` is possible if there's no chunks in the batch. We should verify that `len(b.Chunks) > 0`

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -977,6 +1030,16 @@ func (r *Layer2Relayer) constructCommitBatchPayloadCodecV7(batchesToSubmit []*db
...
 
+		// Check encoding batch fields are not zero hashes
+		if encodingBatch.ParentBatchHash == (common.Hash{}) {
```
jonastheis: Aren't these checks already done when we store the batch to the DB?

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Validate each batch in detail
+	if err := r.validateBatchesDetailed(batchesToSubmit, prevChunk); err != nil {
+		return err
+	}
+
+	log.Info("Sanity check passed before constructing transaction", "batches count", len(batchesToSubmit))
```
jonastheis: Do we need this log? if we submit we can assume that the sanity checks passed. if not we should log an error. we will configure alerting such that we will be notified of error logs

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1017,16 +1105,55 @@ func (r *Layer2Relayer) constructCommitBatchPayloadValidium(batch *dbBatchWithCh
...
+	if len(calldata) == 0 {
+		return nil, 0, 0, fmt.Errorf("generated calldata is empty for batch %d", batch.Batch.Index)
+	}
+
 	log.Info("Validium commitBatch", "maxBlockHeight", maxBlockHeight, "commitment", commitment.Hex())
 	return calldata, maxBlockHeight, totalGasUsed, nil
 }
 
 func (r *Layer2Relayer) constructFinalizeBundlePayloadCodecV7(dbBatch *orm.Batch, endChunk *orm.Chunk, aggProof *message.OpenVMBundleProof) ([]byte, error) {
+	// Basic sanity checks
```
jonastheis: Are all these basic sanity checks really necessary? I'm afraid that they simply bloat the code and make it less readable without having a real impact. Can't we catch all these cases when checking with `sanityChecksBeforeConstructingTransaction`?

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,271 @@
...
+		return err
+	}
+
+	// Validate message queue consistency
+	if err := r.validateMessageQueueConsistency(batch.Batch.Index, batch.Chunks, common.HexToHash(batch.Batch.PrevL1MessageQueueHash), common.HexToHash(batch.Batch.PostL1MessageQueueHash)); err != nil {
+		return err
+	}
+
+	// Validate chunks
+	if err := r.validateBatchChunks(batch, i, allBatches, prevChunkFromPrevBatch); err != nil {
```
jonastheis: why not pass the correct `lastChunkFromPrevBatch` here instead of doing it inside the function?

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	// Get previous chunk for continuity check
+	prevChunk, err := r.getPreviousChunkForContinuity(batchesToSubmit[0])
+	if err != nil {
+		return err
+	}
+
+	// Validate each batch in detail
+	if err := r.validateBatchesDetailed(batchesToSubmit, prevChunk); err != nil {
```
jonastheis: Why not do the checks of `validateBatchesBasic` and `validateCodecVersions` inside `validateBatchesDetailed` (can rename to `validateBatches`)? I think this would make it a bit easier to read.

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -473,6 +491,12 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
 		log.Info("Forcing submission of batches due to timeout", "batch index", batchesToSubmit[0].Batch.Index, "first block created at", oldestBlockTimestamp)
 	}
 
+	// Sanity checks before constructing the transaction
+	if err := r.sanityChecksBeforeConstructingTransaction(batchesToSubmit); err != nil {
```
jonastheis: Some of these checks seem the same (or at least very similar) as they should be done when constructing the `batchesToSubmit`. 

I'm thinking whether it would be more meaningful to do these checks based on the constructed transaction. ie we decode the batch header from calldata and everything else from the blobs and then try to make sense of the decoded data (and relate it back to the DB). This way we would catch any error in constructing the batches and encoding.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: What does the code of this comment point to?

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Sounds good. Combined these small functions in commit: [1b28ff5](https://github.com/scroll-tech/scroll/pull/1714/commits/1b28ff5b5ae43357cd24d27f0eb23a7d4eaefa9a). Previously, this was due to a go lint error. However, making too many small functions seems somewhat overkill.

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Refactored in commit: [1b28ff5](https://github.com/scroll-tech/scroll/pull/1714/commits/1b28ff5b5ae43357cd24d27f0eb23a7d4eaefa9a).

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Removed in commit: [1b28ff5](https://github.com/scroll-tech/scroll/pull/1714/commits/1b28ff5b5ae43357cd24d27f0eb23a7d4eaefa9a).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: No, I think when we store this batch in DB, we didn't check if this hash is accidentally overwritten.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Yeah. It was the first rationale to check everything in `sanityChecksBeforeConstructingTransaction`, but later I found it has many initialization and assignments inside each function, such as `constructFinalizeBundlePayloadCodecV7`, thus adding double-checks in them.

colinlyguo: Removed some of the basic checks in commit: [1b28ff5](https://github.com/scroll-tech/scroll/pull/1714/commits/1b28ff5b5ae43357cd24d27f0eb23a7d4eaefa9a).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Added in commit: [925ef74](https://github.com/scroll-tech/scroll/pull/1714/commits/925ef74867c4a2f3beca276aa3b0c6c834ff4784).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: I see. Removed.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Refactored in commit: [cf85225](https://github.com/scroll-tech/scroll/pull/1714/commits/cf852259f36191d55e7d92a77b8fd8a6523d667d).

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Refactored in commit: [cf85225](https://github.com/scroll-tech/scroll/pull/1714/commits/cf852259f36191d55e7d92a77b8fd8a6523d667d).

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -999,6 +1010,18 @@ func (r *Layer2Relayer) constructCommitBatchPayloadCodecV7(batchesToSubmit []*db
 }
 
 func (r *Layer2Relayer) constructCommitBatchPayloadValidium(batch *dbBatchWithChunks) ([]byte, uint64, uint64, error) {
+	// Check state root is not zero
+	stateRoot := common.HexToHash(batch.Batch.StateRoot)
```
Thegaram, line 1014: Why not also check withdraw root?

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,438 @@
...
+	if firstChunk.Index == 0 {
+		return fmt.Errorf("genesis chunk should not be in normal batch submission flow, chunk index: %d", firstChunk.Index)
+	}
+
+	prevChunk, err := r.chunkOrm.GetChunkByIndex(r.ctx, firstChunk.Index-1)
+	if err != nil {
+		return fmt.Errorf("failed to get previous chunk %d for continuity check: %w", firstChunk.Index-1, err)
+	}
+
+	// Validate codec version consistency across all batches
```
jonastheis: update comment to contain all checks or just remove

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,438 @@
...
+	}
+
+	// Validate block data
+	decodedBlocks := payload.Blocks()
+	if len(decodedBlocks) != len(batchBlocks) {
+		return fmt.Errorf("block count mismatch: decoded=%d, db=%d", len(decodedBlocks), len(batchBlocks))
+	}
+
+	for j, dbBlock := range batchBlocks {
+		if decodedBlocks[j].Number() != dbBlock.Header.Number.Uint64() {
```
jonastheis: Should we also compare the block hash here?

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,438 @@
...
+func (r *Layer2Relayer) validateBlobsAgainstDatabase(blobs []*kzg4844.Blob, batchesToValidate []*dbBatchWithChunks) error {
+	// Get codec for blob decoding
+	firstBatch := batchesToValidate[0].Batch
+	codec, err := encoding.CodecFromVersion(encoding.CodecVersion(firstBatch.CodecVersion))
+	if err != nil {
+		return fmt.Errorf("failed to get codec: %w", err)
+	}
+
+	// Validate each blob against its corresponding batch
+	for i, blob := range blobs {
```
jonastheis, line 339: Not sure if this would be the right place but I think it would be good to also compare the batch hashes of the calldata + blob hash vs database

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -0,0 +1,438 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to decode blob: %w", err)
+	}
+
+	// Validate L1 message queue hashes
+	if payload.PrevL1MessageQueueHash() != common.HexToHash(dbBatch.PrevL1MessageQueueHash) {
+		return fmt.Errorf("prevL1MessageQueueHash mismatch: decoded=%s, db=%s", payload.PrevL1MessageQueueHash().Hex(), dbBatch.PrevL1MessageQueueHash)
+	}
+
+	if payload.PostL1MessageQueueHash() != common.HexToHash(dbBatch.PostL1MessageQueueHash) {
```
jonastheis: Does this catch the case where we wrongly compute the `L1MessageQueueHash` while constructing chunks/batches? I'm afraid that we might just check wrong value against wrong value here

colinlyguo: Yes. This check cannot defend against wrongly computing `L1MessageQueueHash` error, it just checks that the blob payload's message queue hash values are the same as values stored in DB. 

The check which can detect regression error is in function `validateMessageQueueConsistency`: https://github.com/scroll-tech/scroll/blob/cf852259f36191d55e7d92a77b8fd8a6523d667d/rollup/internal/controller/relayer/l2_relayer_sanity.go#L406-L438.

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Thanks. The comment is fixed in https://github.com/scroll-tech/scroll/pull/1714/commits/be22866f4290b185574142c766d10036e45a7df3.

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Added in [3f9d9cc](https://github.com/scroll-tech/scroll/pull/1714/commits/3f9d9cc1b57103344fa1a6e8fb80e1b9bc49f0e6).

rollup/internal/controller/relayer/l2_relayer_sanity.go
colinlyguo: Removed this block-level check after adding batch hash check, since the batch is constructed from blocks: [3f9d9cc](https://github.com/scroll-tech/scroll/pull/1714/commits/3f9d9cc1b57103344fa1a6e8fb80e1b9bc49f0e6).

rollup/internal/controller/relayer/l2_relayer_sanity.go
```diff
@@ -436,3 +397,51 @@ func (r *Layer2Relayer) validateMessageQueueConsistency(batchIndex uint64, chunk
 
 	return nil
 }
+
+func assembleDABatchFromPayload(calldataInfo *CalldataInfo, payload encoding.DABlobPayload, dbBatch *orm.Batch, codec encoding.Codec) (encoding.DABatch, error) {
+	blocks, err := assembleBlocksFromPayload(payload)
+	if err != nil {
+		return nil, fmt.Errorf("failed to assemble blocks from payload batch_index=%d codec_version=%d parent_batch_hash=%s: %w", dbBatch.Index, dbBatch.CodecVersion, calldataInfo.ParentBatchHash.Hex(), err)
+	}
+	parentBatchHash := calldataInfo.ParentBatchHash
```
jonastheis: Does this work with multiple blobs/batches in the same tx? Seems like we need to pass the parent batch hash here from the loop in `validateCalldataAndBlobsAgainstDatabase`

colinlyguo: Nice catch. Fixed in [2d4f792](https://github.com/scroll-tech/scroll/pull/1714/commits/2d4f792a6640113900ce6b6660cb9379b6c3c98f).

rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -1028,6 +1052,17 @@ func (r *Layer2Relayer) constructCommitBatchPayloadValidium(batch *dbBatchWithCh
...
+	// Check state root is not zero
+	stateRoot := common.HexToHash(dbBatch.StateRoot)
+	if stateRoot == (common.Hash{}) {
+		return nil, fmt.Errorf("batch %d state root is zero", dbBatch.Index)
+	}
+
+	// Check proof if present
+	if aggProof != nil && len(aggProof.Proof()) == 0 {
+		return nil, fmt.Errorf("aggregate proof is empty")
+	}
```
Thegaram: Is this useful? Finalization should fail anyway

Thegaram (commented):
lgtm

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: The withdraw root could be empty at first, thus not adding the withdraw root's non-empty check in case of an e2e test case.

rollup/internal/controller/relayer/l2_relayer.go
colinlyguo: Makes sense. And also proof has its sanity checks in [`SanityCheck()`](https://github.com/scroll-tech/scroll/blob/871728190c18ca112a9ac3c2d9d8e8b97a2957e4/common/types/message/message.go#L204), thus removing this check.

jonastheis (dismissed):

jonastheis (approved):

georgehao (approved):

PR #1719, opened 2025-08-09
[Feat] utilities for debugging

noel2004:
1. Handling proving task and generate it into universal task

  + Copy the dumped json file for proving task (can be extract from db or /tmp if ZKVM_DEBUG flag is enabled for coordinator) and naming it as `<chunk/batch/bundle>_proving_task.json`
  + Under `crates/libzkp`, run

~~~ sh
cargo test --release --test test_gen_universal_task_<chunk/batch/bundle> -- --ignored
~~~

```
cargo test --package libzkp --test tasks -- tests::test_gen_universal_task_bundle --exact --show-output --ignored
```

It would generate the universal task and dump it as `dump_univ_task.json` under the same directory. And we can debug locally for any issues encountered in running of tests

### DIFF
--- a/crates/libzkp/.gitignore
+++ b/crates/libzkp/.gitignore
@@ -1,0 +1,1 @@
+testdata/*.json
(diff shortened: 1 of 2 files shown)


lispc:
i feel we also also dump a {chunk,batch,bundle}Witness json.   And in zkvm-prover, i will load this json as input.   At least json is "viewable & editable".

PR #1733, opened 2025-09-03
[Fix] Compatible with prover from 4.5.33

noel2004:
Unexpected serde behavior in OpenVm proof has broken the backward compatibility of coordinator. This PR detect the prover version and apply fixing to the encoding of universal task to resume the compatibility

## Summary by CodeRabbit

- New Features
  - Automatic compatibility handling for universal prover tasks based on prover version; tasks are transformed when required to ensure compatibility.
  - Exposed a compatibility-fix API to convert universal task payloads to legacy-compatible format.

- Bug Fixes
  - Improved error handling and logging during compatibility processing to prevent invalid task dispatch.

- Chores
  - Updated internal dependency revisions for prover, verifier, and types crates.
  - Test environment and package version tags updated to newer prover/version releases.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8395,7 +8395,7 @@ dependencies = [
 [[package]]
 name = "scroll-zkvm-prover"
 version = "0.5.0"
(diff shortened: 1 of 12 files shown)


noel2004:
deprecated by #1742

PR #1735, opened 2025-09-08
feat(gas-oracle): support moving average base fee

yiweichi:

This PR updates the gas-oracle to use a moving average of the base fee, rather than relying solely on the latest block’s baseFee and blobBaseFee. This approach should provide a more accurate representation of the L1 fee.

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes

## Summary by CodeRabbit

- New Features
  - Gas oracle now averages recent L1 base fee and blob base fee for more stable fee estimates.
  - Added configurable averaging window (calculate_average_fees_window_size); default 100.
  - Safer fee computation when an alternative gas token is enabled, with overflow protection and clamping.

- Bug Fixes / Chores
  - Improved handling and logging when no recent blocks are available.
  - Version bumped to v4.5.47.


rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -105,23 +105,20 @@ func NewLayer1Relayer(ctx context.Context, db *gorm.DB, cfg *config.RelayerConfi
...
-		log.Error("Failed to GetL1Blocks from db", "height", latestBlockHeight, "err", err)
+		log.Error("Failed to GetLatestL1Blocks from db", "limit", limit, "err", err)
 		return
 	}
-	if len(blocks) != 1 {
-		log.Error("Block not exist", "height", latestBlockHeight)
+
+	// nothing to do if we don't have any l1 blocks
+	if len(blocks) == 0 {
+		log.Info("No l1 blocks to process", "limit", limit)
```
Thegaram: Let's use `Warn` here since this should normally not happen.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -105,23 +105,20 @@ func NewLayer1Relayer(ctx context.Context, db *gorm.DB, cfg *config.RelayerConfi
...
 		return
 	}
+
 	block := blocks[0]
```
Thegaram, line 122: Is this the oldest or newest block?

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -130,8 +127,8 @@ func (r *Layer1Relayer) ProcessGasPriceOracle() {
 			return
 		}
 
-		baseFee := block.BaseFee
-		blobBaseFee := block.BlobBaseFee
+		// calculate the average base fee and blob base fee of the last N blocks
+		baseFee, blobBaseFee := r.calculateAverageFees(blocks)
```
Thegaram, line 131: Many places below still use `block`.
- Threshold should be checked against `blobBaseFee`, not `block.BlobBaseFee`
- Send tx should probably not use `block.Hash` for ID (risk of conflict with other txs)
- Need to consider DB updates, including gas oracle import status

rollup/internal/orm/l1_block.go
```diff
@@ -71,6 +71,46 @@ func (o *L1Block) GetL1Blocks(ctx context.Context, fields map[string]interface{}
...
+
+	var l1Blocks []L1Block
+	if err := db.Find(&l1Blocks).Error; err != nil {
+		return nil, fmt.Errorf("L1Block.GetLatestL1Blocks error: %w, limit: %d", err, limit)
+	}
+	return l1Blocks, nil
+}
+
+// GetAverageFees returns the average base_fee and blob_base_fee of all blocks
+func (o *L1Block) GetAverageFees(ctx context.Context) (avgBaseFee float64, avgBlobBaseFee float64, err error) {
```
Thegaram: Is this not used?

Thegaram (changes requested):

rollup/internal/controller/relayer/l1_relayer.go
yiweichi: it's the newest block.

rollup/internal/orm/l1_block.go
yiweichi: removed, should be AI generated this.

rollup/internal/controller/relayer/l1_relayer.go
yiweichi: > Send tx should probably not use block.Hash for ID (risk of conflict with other txs)

I think we can add a prefix `updateL1GasOracle-` before `block.Hash`. btw, just want to check this is an already existing risk, instead of introducing by this PR right?

yiweichi: > Need to consider DB updates, including gas oracle import status

I'm not sure what does that mean, I think we already update import status [here](https://github.com/scroll-tech/scroll/blob/de7f6e56a9698e29ebfcc64c354e7ebec75dd315/rollup/internal/controller/relayer/l1_relayer.go#L188) and [here](https://github.com/scroll-tech/scroll/blob/de7f6e56a9698e29ebfcc64c354e7ebec75dd315/rollup/internal/controller/relayer/l1_relayer.go#L217).

Thegaram: Currently we update the status to `GasOracleImporting` and then to `GasOracleImported`, ensuring that we import the same L1 block info at most once.

If your intention that we keep exactly the same logic, but now instead of meaning "this block is being imported / has been imported", it will mean that "the avg price with the window up to this block is being imported / has been imported"?

In that case, my points 2 and 3 above can be ignored. But please double check that this will work as intended.

rollup/internal/orm/l1_block.go
Thegaram: Please always carefully review AI-generated code

rollup/internal/controller/relayer/l1_relayer.go
yiweichi: > it will mean that "the avg price with the window up to this block is being imported / has been imported"?

Yes, that's my intention. We only care about if the newest block of the window has been imported/importing or not. If it has, we just skip it, and waiting for the next new l1 block.

I think this will work as intended.

PR #1742, opened 2025-09-24
[Feat] For prover 4.6.1

noel2004:
This PR is prepared for prover 4.6.1, in which zkvm use openvm 1.4. It has updated the zkvm-prover dependency and cherry-pick some commits from `cloak` branch

It is also ready for coordinator to be deployed in mainnet (compatible with the online prover `4.5.33` enabling us a smooth upgrading)

local e2e test has passed for following configurations:

- [x] coordinator + prover (both built with current PR)  + guest 0.5.2
- [x] coordinator (from `cloak` branch) + prover (built with current PR)  + guest 0.6.0-rc.1
- [x] coordinator (built with current PR) + prover (4.5.33)  + guest 0.5.2

## Summary by CodeRabbit

- New Features
  - Automatic compatibility adjustment for universal prover tasks to support older provers.

- Build/Setup
  - Local setup uses a dedicated config.template path and ensures required directories are created.
  - Release process now downloads an additional verifier asset for improved setup.

- Configuration
  - Default local DB connection updated to include dev credentials.
  - Application version bumped to v4.6.1.

- Tests
  - E2E tests add default block-range variables and convert SQL dumps into migration-friendly files.

- Chores
  - Updated workspace dependency revisions for zkvm components.


coordinator/internal/logic/libzkp/lib.go
```diff
@@ -147,3 +147,20 @@ func SetDynamicFeature(feats string) {
 	defer freeCString(cFeats)
 	C.set_dynamic_feature(cFeats)
 }
+
+// UnivTaskCompatibilityFix calls the universal task compatibility fix function
+func UnivTaskCompatibilityFix(taskJSON string) (string, error) {
```
lispc: full name is always strictly better.  Change all "univ" -> "universal"

crates/libzkp/src/lib.rs
```diff
@@ -46,6 +46,47 @@ pub fn checkout_chunk_task(
     Ok(ret)
 }
 
+/// Convert the universal task json into compatible form for old prover
+pub fn univ_task_compatibility_fix(task_json: &str) -> eyre::Result<String> {
+    use scroll_zkvm_types::proof::VmInternalStarkProof;
+
+    let u_task: tasks::ProvingTask = serde_json::from_str(task_json)?;
```
lispc: either "univerial_task" or just "task".

Cargo.toml
```diff
@@ -17,9 +17,10 @@ repository = "https://github.com/scroll-tech/scroll"
 version = "4.5.47"
 
 [workspace.dependencies]
-scroll-zkvm-prover = { git = "https://github.com/scroll-tech/zkvm-prover", rev = "060be4c" }
-scroll-zkvm-verifier = { git = "https://github.com/scroll-tech/zkvm-prover", rev = "060be4c" }
-scroll-zkvm-types = { git = "https://github.com/scroll-tech/zkvm-prover", rev = "060be4c" }
+# include compatiblity fixing from "fix/coordinator"
+scroll-zkvm-prover = { git = "https://github.com/scroll-tech/zkvm-prover", rev = "2c7fd3c" }
```
lispc: is this a commit in a branch or master?

noel2004: A branch: https://github.com/scroll-tech/zkvm-prover/tree/compatibility-4_5_33, base on master with tirival update (to expose the proof type for serialize)

lispc (dismissed):

coordinator/conf/config.json
```diff
@@ -22,7 +22,7 @@
   },
   "db": {
     "driver_name": "postgres",
-    "dsn": "postgres://localhost/scroll?sslmode=disable",
+    "dsn": "postgres://dev:dev@localhost/scroll?sslmode=disable",
```
georgehao: don't change this

crates/libzkp/src/lib.rs
```diff
@@ -46,6 +46,47 @@ pub fn checkout_chunk_task(
     Ok(ret)
 }
 
+/// Convert the universal task json into compatible form for old prover
+pub fn univ_task_compatibility_fix(task_json: &str) -> eyre::Result<String> {
```
georgehao, line 50: This function transfers the original structure to a new structure. I don't like this way. How about directly implementing it in go?

noel2004: 1. We need to convert the structure with some internal types defined in openvm so it would be rather hard to imply them by Go.
2. This is a temporary resolution only for upgrade our old prover (base on 4.5.33) and will not be used again later. So it may not bother to spend more effort on that.

georgehao (approved):

lispc (approved):

PR #1745, opened 2025-10-02
feat(l2 relayer): enhance batch submission strategy based on backlog size

jonastheis:
Adds more dynamic blob submission.
```
backlogCount >= backlog_max: submit min_batches for fast inclusion at slightly higher price.
backlogCount < backlog_max: submit max_batches for better cost amortization.
```

### PR title

- [x] feat: A new feature
- [ ] fix: A bug fix

### Deployment tag versioning

- [x] Yes

### Breaking change label

- [ ] Yes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved batch submission logic to be backlog-aware, reducing overload during high traffic and enhancing stability and throughput consistency.

* **Chores**
  * Bumped version to v4.5.48.

### DIFF
--- a/rollup/internal/controller/relayer/l2_relayer.go
+++ b/rollup/internal/controller/relayer/l2_relayer.go
@@ -345,8 +345,16 @@ func (r *Layer2Relayer) commitGenesisBatch(batchHash string, batchHeader []byte,
 // - backlogCount > r.cfg.BatchSubmission.BacklogMax -> forceSubmit
 // - we have at least minBatches AND price hits a desired target price
 func (r *Layer2Relayer) ProcessPendingBatches() {
-	// Get effective batch limits based on whether validium mode is enabled.
-	minBatches, maxBatches := r.getEffectiveBatchLimits()
+	// First, get the backlog count to determine batch submission strategy
+	backlogCount, err := r.batchOrm.GetFailedAndPendingBatchesCount(r.ctx)
+	if err != nil {
+		log.Error("Failed to fetch pending L2 batches count", "err", err)
+		return
+	}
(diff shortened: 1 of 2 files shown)


rollup/internal/controller/relayer/l2_relayer.go
```diff
@@ -563,12 +562,22 @@ func (r *Layer2Relayer) ProcessPendingBatches() {
...
 	}
-	return r.cfg.BatchSubmission.MinBatches, r.cfg.BatchSubmission.MaxBatches
+
+	// If backlog is at or above max, prioritize fast inclusion by submitting min_batches
+	if backlogCount >= r.cfg.BatchSubmission.BacklogMax {
+		return r.cfg.BatchSubmission.MinBatches, r.cfg.BatchSubmission.MinBatches
+	}
+
+	// Otherwise, prioritize cost efficiency by trying to submit max_batches
+	return r.cfg.BatchSubmission.MaxBatches, r.cfg.BatchSubmission.MaxBatches
```
Thegaram, line 580: Should this be `MinBatches, MaxBatches`?

Either way seems fine to me. It only applies to the case when we have very low batch creation rate.

Thegaram (approved):

rollup/internal/controller/relayer/l2_relayer.go
jonastheis: No I think max_batches here makes sense. Otherwise we would already commit if we reach min_batches even if the catch creation rate is low. In practice this would mean we always commit min_batches. 

But here we want to force to wait until we reach max_batches instead (amortize more) at the cost of slower committing.

Either way we will commit if we reach the timeout

georgehao (approved):

PR #1753, opened 2025-11-18
[Feat] Galileo forking

noel2004:
This PR induced works required by galileo forking

- [x] Update zkvm-prover for galileo forking
- [x] Update libzkp to handle galileo forking task
- [x] Improve feature settings to dyanmic support runtime options
- [x] Stuffs for local e2e tests
- [x] e2e tests (for galileo blocks and `v0.7.0-rc` guests
regression e2e tests: 
   - [x] for feynman blocks and `v0.5.2` guests
   - [x] for feynman blocks, with prover `4.6.1` and  `v0.5.2` guests

## Summary by CodeRabbit

* **New Features**
  * Release bumped to v4.7.1; per-circuit/per-task feature flags (e.g., OpenVM 13) and multi-version/multi-codec proof handling; added test network genesis/config templates.

* **Bug Fixes**
  * Now errors on unsupported codec versions; stronger batch/header/version consistency checks and clearer informational logging.

* **Improvements**
  * Dependency references aligned to fixed tags; circuit configs accept optional feature strings; consolidated task precheck/serialization flows.

* **Chores**
  * Test infra: new Make targets, DB reset/reimport flows, env defaults, and setup filtering.


common/version/version.go
```diff
@@ -5,7 +5,7 @@ import (
 	"runtime/debug"
 )
 
-var tag = "v4.6.3"
+var tag = "v4.7.1"
```
Thegaram, line 8: Why not `4.7.0`?

crates/libzkp/src/tasks/batch.rs
```diff
@@ -26,23 +26,25 @@ pub struct BatchHeaderValidiumWithHash {
...
-/// Define variable batch header type, since BatchHeaderV6 can not
-/// be decoded as V7 we can always has correct deserialization
-/// Notice: V6 header MUST be put above V7 since untagged enum
-/// try to decode each defination in order
+/// Parse header types passed from golang side and adapt to the
+/// defination in zkvm-prover's types
+/// We distinguish the header type in golang side according to the codec
+/// version, i.e. v6 - v9 (current), and validium
+/// And adapt it to different header version used in zkvm-prover's witness
+/// defination, i.e. v6- v8 (current), and validium
```
Thegaram:
```suggestion
/// definition in zkvm-prover's types
/// We distinguish the header type in golang side according to the codec
/// version, i.e. v7 - v9 (current), and validium
/// And adapt it to the corresponding header version used in zkvm-prover's witness
/// definition, i.e. v7 - v8 (current), and validium
```

crates/libzkp/src/tasks/batch.rs
```diff
@@ -139,51 +141,55 @@ impl TryFrom<BatchProvingTask> for ProvingTask {
...
-                                "hardfork mismatch for da-codec@v7 header: found={}, expected={:?}",
-                                fork_name,
-                                [ForkName::EuclidV2, ForkName::Feynman],
-                            ),
-                        }
-                    }
-                    BatchHeaderV::Validium(_) => unreachable!("version!=validium"),
+                    Codec::V7 => <EnvelopeV7 as Envelope>::from_slice(padded_blob_bytes.as_slice())
+                        .challenge_digest(versioned_hash),
+                    Codec::V8 => <EnvelopeV8 as Envelope>::from_slice(padded_blob_bytes.as_slice())
```
Thegaram: What about V9?

crates/libzkp/src/tasks/batch.rs
```diff
@@ -234,7 +240,7 @@ impl BatchProvingTask {
             (Domain::Scroll, STFVersion::V7) => {
                 ReferenceHeader::V7(*self.batch_header.must_v7_header())
             }
-            (Domain::Scroll, STFVersion::V8) => {
+            (Domain::Scroll, STFVersion::V8) | (Domain::Scroll, STFVersion::V9) => {
                 ReferenceHeader::V8(*self.batch_header.must_v8_header())
```
Thegaram: V9 uses v9 header, no?

crates/libzkp/src/tasks/batch.rs
```diff
@@ -139,51 +141,55 @@ impl TryFrom<BatchProvingTask> for ProvingTask {
...
-                                fork_name,
-                                [ForkName::EuclidV2, ForkName::Feynman],
-                            ),
-                        }
-                    }
-                    BatchHeaderV::Validium(_) => unreachable!("version!=validium"),
+                    Codec::V7 => <EnvelopeV7 as Envelope>::from_slice(padded_blob_bytes.as_slice())
+                        .challenge_digest(versioned_hash),
+                    Codec::V8 => <EnvelopeV8 as Envelope>::from_slice(padded_blob_bytes.as_slice())
+                        .challenge_digest(versioned_hash),
```
georgehao: don't need the Codec::V9?

crates/prover-bin/src/zk_circuits_handler/universal.rs
```diff
@@ -3,27 +3,34 @@ use std::path::Path;
...
 use scroll_zkvm_prover::{Prover, ProverConfig};
 use scroll_zkvm_types::ProvingTask;
 use tokio::sync::Mutex;
 pub struct UniversalHandler {
     prover: Prover,
 }
 
+// additional config dispatched with proving task
+#[derive(Debug, Default)]
+pub(crate) struct TaskConfig {
```
georgehao: very wierd, I think ProvintTaskExt is enough, don't need to have a new TaskConfig

crates/libzkp/src/tasks/batch.rs
roynalnaruto: Since the `zkvm-prover` has introduced the `Version` that includes both `stf_version` and `codec`, on our side we stopped adding new variants for `Codec` for every upgrade.

So `Codec::V8` is OK here, from the perspective of `zkvm-prover`. The implementation between `da-codec` and `zkvm-prover` is different, but `zkvm-prover` makes the `version` byte comparison against `stf_version`, which _is_ in fact `9` for `Galileo`.

crates/libzkp/src/tasks/batch.rs
roynalnaruto: Similarly, `ReferenceHeader::V8` is identical to `Galileo`'s header so in `zkvm-prover` we avoided adding more variants. Instead we do the matching on `stf_version`, on L#243.

noel2004:
e2e test (for galileo blocks with guest `v0.7.0-rc.3`) passed.

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -311,9 +311,12 @@ func (bp *BatchProverTask) getBatchTaskDetail(dbBatch *orm.Batch, chunkProofs []
 	if !bp.validiumMode() {
 		dbBatchCodecVersion := encoding.CodecVersion(dbBatch.CodecVersion)
 		switch dbBatchCodecVersion {
+		case 0:
+			log.Warn("the codec version is 0, if it is not under integration test we have encountered an error here")
```
roynalnaruto, line 315: Does this mean we use codec=0 for integration tests?

crates/libzkp/src/lib.rs
```diff
@@ -1,38 +1,64 @@
...
-            s => tracing::warn!("unrecognized dynamic feature: {s}"),
         }
+        ret
     }
 }
 
+pub fn set_dynamic_feature(_: &str) {
+    tracing::error!("call deprecated method");
+}
+
```
roynalnaruto: can we just remove this from `libzkp`? And all occurrences of it from the scroll monorepo?

crates/libzkp/src/lib.rs
```diff
@@ -1,38 +1,64 @@
...
-        match feat_s.trim().to_lowercase().as_str() {
-            "legacy_witness" => {
-                tracing::info!("set witness encoding for legacy mode");
-                unsafe {
-                    // the function is only called while initialize step
-                    LEGACY_WITNESS_ENCODING = true;
+static ADDITIONAL_FEATURES: OnceLock<HashMap<HardForkName, FeatureOptions>> = OnceLock::new();
+
+impl FeatureOptions {
+    pub fn new(feats: &str) -> Self {
```
roynalnaruto, line 38: Why not instead:

```rust
impl FromStr for FeatureOptions
```

crates/libzkp/src/lib.rs
```diff
@@ -1,38 +1,64 @@
...
+                "can not found features setting for unrecognized fork {}",
+                fork_name
+            )
+        })
+}
 
-pub(crate) fn witness_use_legacy_mode() -> bool {
-    unsafe { LEGACY_WITNESS_ENCODING }
+#[derive(Debug, Default, Clone)]
+struct FeatureOptions {
```
roynalnaruto, line 30: If I'm not wrong: `legacy_witness <=> openvm_13` ?

crates/libzkp/src/tasks.rs
```diff
@@ -35,17 +35,37 @@ fn check_aggregation_proofs<Metadata: MultiVersionPublicInputs>(
     Ok(())
 }
 
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct ProvintTaskExt {
```
roynalnaruto: Maybe you mean `ProvingTaskExt` ? We should fix this typo

crates/prover-bin/src/zk_circuits_handler/universal.rs
```diff
@@ -36,8 +43,13 @@ impl UniversalHandler {
         &mut self.prover
     }
 
-    pub fn get_task_from_input(input: &str) -> Result<ProvingTask> {
-        Ok(serde_json::from_str(input)?)
+    pub fn get_task_from_input(input: &str) -> Result<(ProvingTask, TaskConfig)> {
```
roynalnaruto: Why not
```suggestion
    pub fn get_task_from_input(input: &str) -> Result<ProvingTaskExt> {
```

coordinator/internal/logic/provertask/batch_prover_task.go
Thegaram: ~I think it should never be 0~ We only use v0 for genesis batch (which is not proven)

coordinator/internal/logic/provertask/batch_prover_task.go
```diff
@@ -325,6 +328,7 @@ func (bp *BatchProverTask) getBatchTaskDetail(dbBatch *orm.Batch, chunkProofs []
 		if decodeErr != nil {
 			return nil, fmt.Errorf("failed to decode batch header version %d: %w", dbBatch.CodecVersion, decodeErr)
 		}
+		log.Info("Decode batchheader bytes to canonical header", "version", batchHeader.Version())
```
Thegaram: Do we need this log?

crates/libzkp/src/lib.rs
```diff
@@ -1,38 +1,64 @@
...
-/// global features: use legacy encoding for witness
-static mut LEGACY_WITNESS_ENCODING: bool = false;
+pub(crate) fn witness_use_legacy_mode(fork_name: &str) -> eyre::Result<bool> {
+    ADDITIONAL_FEATURES
+        .get()
+        .and_then(|features| features.get(fork_name))
+        .map(|cfg| cfg.legacy_witness_encoding)
+        .ok_or_else(|| {
+            eyre::eyre!(
+                "can not found features setting for unrecognized fork {}",
```
Thegaram: Do you mean "cannot find feature settings"?

tests/prover-e2e/cloak-xen/config.template.json
```diff
@@ -0,0 +1,40 @@
...
+      "endpoint": "http://cloak-xen-sequencer.sepolia.scroll.tech:8545/"
+    }
+  },
+  "auth": {
+    "secret": "prover secret key",
+    "challenge_expire_duration_sec": 3600,
+    "login_expire_duration_sec": 3600
+  },
+  "sequencer": {
+    "decryption_key": "ad25ba5de9f03fd9e3c81a1ad731c5c7e94ae2641f75ed7d3275312053b40f75"
```
Thegaram: Why do you commit a public network's secret into GitHub.....

Thegaram (commented):
lgtm

tests/prover-e2e/cloak-xen/config.template.json
noel2004: Very sorry! It is the secrect for testnet and put into commit by mistake...

I will remove it in the next commit but of course we have to deprecate this secrect if it is also used by public network...

crates/libzkp/src/lib.rs
noel2004: fixed

coordinator/internal/logic/provertask/batch_prover_task.go
noel2004: In test for Galileo forking, it is used to detect the possible mismatch in batch header encoding.

It would be nonsense after the code has be fixed, removed.

tests/prover-e2e/cloak-xen/config.template.json
Thegaram: Yeah, will do that

common/version/version.go
noel2004: It make sense but since we have buit several images for coordinator / prover base on this version number. So I think it would better not to rewind it now.

crates/libzkp/src/tasks/batch.rs
noel2004: Fixed. We in fact also distinguish v6 for euclidV1 (nothing would touch it any more though).

coordinator/internal/logic/provertask/batch_prover_task.go
noel2004: In integration tests we only provided a dummy task with many fields unfilled, so the version field is 0.

And in integration test, the blob bytes are empty and can not be parsed by any codec, we need to return early here.

It is not a trivial task to refine the whole integration tests so currently we would better just keep this workaround for compatibility.

crates/libzkp/src/lib.rs
noel2004: Not always. `openvm_13` in fact indicate using `k=24` for generating snark proof, including the cloak network which adopt 0.6.0 guests, which has no `legacy_witness` flag.

crates/libzkp/src/lib.rs
noel2004: Now we just tolerate any wrong feature strings so the syntax of try in `FromStr` is not applied.
Not quite a trivial updating so maybe we would better delay this refining.

crates/libzkp/src/lib.rs
noel2004: Purged

georgehao (approved):

lispc (approved):

PR #1755, opened 2025-11-18
(libzkp): refactor batch header variants

roynalnaruto:

Incorporate refactoring changes proposed in https://github.com/scroll-tech/zkvm-prover/pull/225.

This in addition to https://github.com/scroll-tech/scroll/pull/1754 will avoid implementation confusion between `da-codec` and `zkvm-prover`. Since da-codec v7, v8 and v9 are effectively identical (except the version byte), we cover all the 3 cases in a single umbrella variant `V7_V8_V9` and instead make use of the [`Version`](https://github.com/scroll-tech/scroll/blob/feat/galileo/coordinator/internal/utils/version.go) to convey stf-version and codec-version.

Before merging, we must wait for:
- [x] https://github.com/scroll-tech/zkvm-prover/pull/225
- [x] update `zkvm-prover` tag (`rc.4`) instead of depending on `chore/galileo-refactor-header`

### Breaking change label

- [ ] Yes


crates/libzkp/src/tasks/batch.rs
```diff
@@ -237,11 +227,13 @@ impl BatchProvingTask {
             (Domain::Scroll, STFVersion::V6) => {
                 ReferenceHeader::V6(*self.batch_header.must_v6_header())
             }
-            (Domain::Scroll, STFVersion::V7) => {
-                ReferenceHeader::V7(*self.batch_header.must_v7_header())
-            }
-            (Domain::Scroll, STFVersion::V8) | (Domain::Scroll, STFVersion::V9) => {
-                ReferenceHeader::V8(*self.batch_header.must_v8_header())
+            (Domain::Scroll, stf_version)
```
Thegaram: I think you can write this, too, if you want.

```rust
(Domain::Scroll, STFVersion::V7 | STFVersion::V8 | STFVersion::V9)
```

Cargo.toml
```diff
@@ -17,9 +17,9 @@ repository = "https://github.com/scroll-tech/scroll"
 version = "4.7.1"
 
 [workspace.dependencies]
-scroll-zkvm-prover = { git = "https://github.com/scroll-tech/zkvm-prover", branch = "feat/galileo" }
-scroll-zkvm-verifier = { git = "https://github.com/scroll-tech/zkvm-prover", branch = "feat/galileo" }
-scroll-zkvm-types = { git = "https://github.com/scroll-tech/zkvm-prover", branch = "feat/galileo" }
+scroll-zkvm-prover = { git = "https://github.com/scroll-tech/zkvm-prover", branch = "chore/galileo-refactor-header" }
```
Thegaram: I guess this is temporary?

Thegaram (approved):
lgtm

noel2004 (approved):

Cargo.toml
roynalnaruto: Yes. Waiting for some e2e tests to pass on `zkvm-prover` and will then update to `tag = "..."`

crates/libzkp/src/tasks/batch.rs
roynalnaruto: yes, clippy prefers that as well

crates/libzkp/src/tasks.rs
```diff
@@ -79,29 +78,22 @@ pub fn gen_universal_chunk_task(
...
+    let (batch_info, batch_pi_hash) = task.precheck_and_build_metadata()?;
     let proving_task = task.try_into()?;
-
     Ok((
-        batch_hash,
-        BatchProofMetadata {
-            batch_info,
-            batch_hash,
-        },
+        batch_pi_hash,
```
Thegaram, line 85: Oh, the previous `batch_hash` actually meant `batch_pi_hash`? That's confusing... Much clearer now.

Thegaram (approved):

crates/libzkp/src/tasks.rs
roynalnaruto: It was a mistake I think, and it never affected in any way because that value was only used for naming the batch proof file.

But still, its ideal to be fixed. The `batch_hash` (batch header's hash) is already available in `BatchInfo` so I removed the `batch_hash` field altogether from `BatchProofMetadata` struct.

noel2004 (approved):

PR #1764, opened 2025-11-25
[Fix] Accept proof submission even it has been timeout

noel2004:
We faced a delimma in `collection_time` setting (i.e. the time limit to set an assigned prove task being timeout): value too small would cause a possible time consuming task can not be completed since all submission would be rejected by timeout; in the other hand, a too big timeout would take too long to re-assign a task if the connection with assignment (prover) lost.

This PR advise to accept the proof submission even it has been timeout: there is no proper reason to reject the result if it can be verified. With the fixing we can reduce the interval of reassignment without worring about a permanent failure of occasional long-running task. The counter of timeout failure would still be counted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling: timed submissions are now processed through validation, tracked, logged, and their status synchronized instead of being rejected.
  * Added last-chance recovery to ensure assigned prover tasks are retried or reconciled on error.
  * Synchronized prover task status updates to prevent mismatches and lost state during processing.

* **New Files / Configurations**
  * Added testnet/prover end-to-end configs, genesis/state files, and new prover runtime/configuration manifest files.


coordinator/internal/logic/submitproof/proof_receiver.go
```diff
@@ -172,6 +172,18 @@ func (m *ProofReceiverLogic) HandleZkProof(ctx *gin.Context, proofParameter coor
 		return ErrValidatorFailureProverTaskEmpty
 	}
 
+	defer func() {
+		if rerr != nil && types.ProverProveStatus(proverTask.ProvingStatus) == types.ProverAssigned {
```
georgehao, line 176: I don't understand why you added this since [here](https://github.com/scroll-tech/scroll/blob/055f76e6ebe739d64e1bf02b4ce3ff2958972e04/coordinator/internal/logic/submitproof/proof_receiver.go#L290-L295) already done this.

coordinator/internal/logic/submitproof/proof_receiver.go
```diff
@@ -311,6 +323,20 @@ func (m *ProofReceiverLogic) validator(ctx context.Context, proverTask *orm.Prov
 		}
 	}()
 
+	// Internally we overide the timeout failure:
```
georgehao, line 326: This will affect a potential issue in the future upgrade e2e test, as we can't easily find the hard-to-prove tasks. I think this feature is ok, but do we really need this feature

coordinator/internal/logic/submitproof/proof_receiver.go
noel2004: There are many return routines in this function while not all of them ensure the status would be set.

Moreover, there is possiblity that the routine would panic, cause a 500 being returned to prover and prover would keep submitting the proof -> nothing is updated -> keep panic

georgehao: > There are many return routines in this function while not all of them ensure the status would be set.

I don't think so. if have issue now, you must change something wrong.

> here is possiblity that the routine would panic, cause a 500 being ret

it will never happened in the gin http framework. [router.Use(gin.Recovery())](https://github.com/scroll-tech/scroll/blob/develop/coordinator/internal/route/route.go#L16) will capture the panic

coordinator/internal/logic/submitproof/proof_receiver.go
noel2004: Now a chunk task can not be reset within 2hr (7200s).
However we can not make the timeout shorten in case a chunk really needs hours to be proven.
So the feature should be important, @lispc also agree with that.

[This commit](https://github.com/scroll-tech/scroll/pull/1764/commits/1e8507073a72e91ea3e09e3f028a4ec698546f68) try to provide a resolution: we always keep the failure type it is updated before even when the proof is valid. so:

1. For most case, while close task the failure type is `undefined` and nothing change
2. For timeout case, the proof is accepted and task status is updated to be valid, while failure type is still "timeout" and we can still pick up hard tasks by filtering the failure type
3. In rare case, the proof is submitted after timeout, but still failed. In this case, the final failure reason (like verification failure) is more important and is recorded

coordinator/internal/logic/submitproof/proof_receiver.go
noel2004: > it will never happened in the gin http framework. [router.Use(gin.Recovery())](https://github.com/scroll-tech/scroll/blob/develop/coordinator/internal/route/route.go#L16) will capture the panic

This `defer()` call It is not about catching panic. What I mean: a panic cause the routine exit without any updating in db, and return prover a 500 error. So prover would submit this proof again. And since nothing is updated, the same panic would raise again and we have a inifinite loop.

With this `defer` call we always have some progress left in db and it would stop the looping.

> I don't think so. if have issue now, you must change something wrong.

In fact there are. Like [here](https://github.com/scroll-tech/scroll/blob/055f76e6ebe739d64e1bf02b4ce3ff2958972e04/coordinator/internal/logic/submitproof/proof_receiver.go#L252), a malformed proof can not be unmarshalled cause the routine return without update anything.

You can add a `proofRecover` call before any returns but a defer would be more tidy and clear. And it would be needed in panic case, as mentioned above.

georgehao: oh, get it. make sense

georgehao (dismissed):

lispc (dismissed):

georgehao (approved):

lispc (approved):

PR #1772, opened 2025-11-27
fix: Fix blob fee overflow on rollup-relayer and gas-oracle

Thegaram:

Previously we used the l2geth Go SDK's `misc.CalcBlobFee` function to compute the blob base fee corresponding to a header. However, the internal parameters of `CalcBlobFee` have changed in Osaka and subsequent BPO forks. L2geth did not adopt these parameter updates, and as a result, we computed extermely high blob base fee values, which in turn resulted in u64 overflows, erroneous gas oracle updates, and L1 fee estimation failures.

This PR makes the following changes:
- Instead of using `misc.CalcBlobFee`, we will now call the L1 node's `eth_blobBaseFee` API.
- Expose "raw RPC client" to watcher and relayer so that they can call `eth_blobBaseFee`. The "wrapped client" from l2geth does not support this API.
- Add two new `gas-oracle` configurations (`l1_base_fee_limit` and `l1_blob_base_fee_limit`) to better prevent erroneous gas oracle updates in the future.

### PR title

- [X] fix: A bug fix

### Deployment tag versioning

- [X] Yes

### Breaking change label

**Rollout requires config update.**

- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Configurable upper limits for Layer 1 gas prices to cap excessive fees.
  * Blob base fees now fetched from the network for more accurate pricing.
  * New metric tracking when gas fees exceed configured limits.

* **Bug Fixes**
  * Improved error diagnostics in gas estimation with richer payload details.

* **Chores**
  * Version bumped to v4.7.6.
  * Dependency and Docker base image upgrades.


common/testcontainers/testcontainers.go
```diff
@@ -167,13 +167,17 @@ func (t *TestcontainerApps) GetPoSL1EndPoint() (string, error) {
...
-		return nil, err
+		return nil, nil, err
+	}
+	rpcClient, err := rpc.Dial(endpoint)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to dial L1 rpc endpoint, endpoint: %s, err: %w", endpoint, err)
 	}
-	return ethclient.Dial(endpoint)
+	return rpcClient, ethclient.NewClient(rpcClient), nil
 }
```
frisitano: optional nit:
```go
func (t *TestcontainerApps) GetPoSL1Client() (*rpc.Client, *ethclient.Client, error)
```

could be

```go
func (t *TestcontainerApps) GetPoSL1Client() (*rpc.Client, error)
```

and the caller can construct `ethclient.Client` if they need it.

rollup/internal/controller/relayer/l1_relayer.go
```diff
@@ -173,6 +173,18 @@ func (r *Layer1Relayer) ProcessGasPriceOracle() {
 			} else if err != nil {
 				return
 			}
+			// Cap base fee update at the configured upper limit
+			if limit := r.cfg.GasOracleConfig.L1BaseFeeLimit; limit > 0 && baseFee > limit {
```
frisitano: Setting `limit = 0` bypasses this check. Is this intentional?

rollup/internal/config/relayer.go
```diff
@@ -109,6 +109,10 @@ type GasOracleConfig struct {
 	L1BaseFeeDefault                   uint64 `json:"l1_base_fee_default"`
 	L1BlobBaseFeeDefault               uint64 `json:"l1_blob_base_fee_default"`
 
+	// Upper limit values for gas oracle updates
+	L1BaseFeeLimit     uint64 `json:"l1_base_fee_limit"`
+	L1BlobBaseFeeLimit uint64 `json:"l1_blob_base_fee_limit"`
```
frisitano, line 114: Is there a way to validate that `L1BaseFeeLimit > 0` and  `L1BlobBaseFeeLimit > 0`

frisitano (dismissed):
Looks good, left some comments inline.

yiweichi (dismissed):

rollup/internal/config/relayer.go
jonastheis: in rollup relayer we do this kind of validation in the app file: https://github.com/scroll-tech/scroll/blob/6bee33036fa58f8578feae6eeff12a997dcc2d7e/rollup/cmd/rollup_relayer/app/app.go#L85

common/testcontainers/testcontainers.go
Thegaram: Done in [d4948d5](https://github.com/scroll-tech/scroll/pull/1772/commits/d4948d58766cfb2a5fcce2a0b35e120edb006626).

rollup/internal/controller/relayer/l1_relayer.go
Thegaram: Done in [9be56ff](https://github.com/scroll-tech/scroll/pull/1772/commits/9be56ff97d2ab90798a4296486073987dd8e0700).

rollup/internal/config/relayer.go
Thegaram: Done in [9be56ff](https://github.com/scroll-tech/scroll/pull/1772/commits/9be56ff97d2ab90798a4296486073987dd8e0700).

